Issue #9, opened 2026-03-22
In Section 3.5: How is retransmission realized?


gorryfair:
The current text says:
  "Both SCONE Network Elements and application endpoints should support retransmission or periodic re-sending of SCONE packets to ensure reliable delivery."

- Please say more about how a network element perform retransmission of the SCONE Packets? What specifically is retransmitted and how?

mirjak:
This section needs to be removed as it does not comply with the protocol spec

smishra1200:
> "Both SCONE Network Elements and application endpoints should support retransmission or periodic re-sending of SCONE
packets to ensure reliable delivery."

> Please say more about how a network element perform retransmission of the SCONE Packets? What specifically is retransmitted and how?

We addressed this specific point in PR29. The PR covers the point that  a Network Element cannot perform transport-layer retransmissions because it cannot synthesize or inject new SCONE packets. Instead, it relies on the application endpoints to send the packets periodically. There are additional comments from Mirja and Marcus that have to be considered so final text will likely be tweaked.

PR #29, opened 2026-04-18
Update "Retransmission of Rate of Advised Bit-Rate Section"

smishra1200:
GitHub #9. Reworked entire section to align with SCONE Protocol and also expanded on text with trade-off on SCONE NE CPU processing load vs SCONE packet update frequency

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -128,28 +128,19 @@ UDP 4-tuple. Such hints prevent unnecessary default rate-limiting, allow the
 network to signal the maximum allowable bit rate, and reduce CPU
 overhead by eliminating additional classification steps.
 
-## Retransmission of Advised Bit-Rate
-Packet loss or non-delivery of SCONE advice reduces its effectiveness. Because
-Network Elements cannot synthesize or inject new SCONE packets into a flow, they
-cannot perform transport-layer retransmissions (see Section 7.1 of
-{{I-D.ietf-scone-protocol}}). Instead, the reliable delivery of throughput advice
-relies on the periodic sending of SCONE packets by the application endpoints.
-
-To ensure advice is not lost and does not expire, application endpoints are expected
-to send SCONE packets periodically, typically at least twice per monitoring period or
-every 20 to 30 seconds. The SCONE Network Element then takes these opportunities to update
-the throughput advice in the traversing packets. While endpoints could utilize QUIC
-transport-layer acknowledgments to detect lost datagrams and trigger a retransmission
-of a SCONE packet, it is generally simpler and highly effective to achieve reliability
-by simply sending SCONE packets more frequently (see Section 8.1 of {{I-D.ietf-scone-protocol}}).
-
-If the CPU processing load in network elements for updating SCONE packets is not the
-limiting factor, it is preferred for endpoints to send, and network elements to update,
-SCONE packets with a higher frequency than every 20 to 30 seconds to minimize the delay
-in activating the advised bit-rate due to packet loss. However, if the CPU processing
-load in a network element becomes a bottleneck, then the SCONE packet update frequency
-can be scaled back to 20 to 30 seconds. Essentially, there is a trade-off between the
-delay in activating the advised bit-rate due to packet loss versus the CPU processing load.
+## Mitigating Packet Loss
+Packet loss or non-delivery of SCONE advice directly reduces its effectiveness.
+Because the reliable delivery of throughput advice relies entirely on the periodic
+sending of SCONE packets by application endpoints, the network element needs to make
+independent operational decisions on how frequently to update those traversing packets.
+This decision relies on operational considerations such as CPU load and the nature of
+the network policies. A network enforcing dynamic policies might prioritize updating SCONE
+packets immediately upon a policy trigger to minimize the application's reaction time to the
+new limit. Conversely, a network enforcing fixed, subscription-based policies can safely
+scale back its update frequency to conserve CPU resources, provided it still updates SCONE
+packets periodically (e.g., every 20 to 30 seconds). This periodic update frequency ensures
+that the throughput advice reliably reaches the endpoint and does not inadvertently expire
+across the standard monitoring period due to normal packet loss.
 
 ## Frequency of Updates
 The rate at which SCONE updates are issued depends on flow
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -129,10 +129,27 @@ network to signal the maximum allowable bit rate, and reduce CPU
 overhead by eliminating additional classification steps.
 
 ## Retransmission of Advised Bit-Rate
-Packet loss or non-delivery of SCONE advice reduces its effectiveness. Both
-SCONE Network Elements and application endpoints should support retransmission or
-periodic re-sending of SCONE packets to ensure reliable delivery.
-Conformance depends on the behavior of both network and application endpoint.
+Packet loss or non-delivery of SCONE advice reduces its effectiveness. Because
+Network Elements cannot synthesize or inject new SCONE packets into a flow, they
+cannot perform transport-layer retransmissions (see Section 7.1 of
+{{I-D.ietf-scone-protocol}}). Instead, the reliable delivery of throughput advice
+relies on the periodic sending of SCONE packets by the application endpoints.
+
+To ensure advice is not lost and does not expire, application endpoints are expected
+to send SCONE packets periodically, typically at least twice per monitoring period or
+every 20 to 30 seconds. The SCONE Network Element then takes these opportunities to update
+the throughput advice in the traversing packets. While endpoints could utilize QUIC
+transport-layer acknowledgments to detect lost datagrams and trigger a retransmission
+of a SCONE packet, it is generally simpler and highly effective to achieve reliability
+by simply sending SCONE packets more frequently (see Section 8.1 of {{I-D.ietf-scone-protocol}}).
+
+If the CPU processing load in network elements for updating SCONE packets is not the
+limiting factor, it is preferred for endpoints to send, and network elements to update,
+SCONE packets with a higher frequency than every 20 to 30 seconds to minimize the delay
+in activating the advised bit-rate due to packet loss. However, if the CPU processing
+load in a network element becomes a bottleneck, then the SCONE packet update frequency
+can be scaled back to 20 to 30 seconds. Essentially, there is a trade-off between the
+delay in activating the advised bit-rate due to packet loss versus the CPU processing load.
 
 ## Frequency of Updates
 The rate at which SCONE updates are issued depends on flow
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -129,10 +129,27 @@ network to signal the maximum allowable bit rate, and reduce CPU
...
+To ensure advice is not lost and does not expire, application endpoints are expected
+to send SCONE packets periodically, typically at least twice per monitoring period or
+every 20 to 30 seconds. The SCONE Network Element then takes these opportunities to update
+the throughput advice in the traversing packets. While endpoints could utilize QUIC
+transport-layer acknowledgments to detect lost datagrams and trigger a retransmission
+of a SCONE packet, it is generally simpler and highly effective to achieve reliability
+by simply sending SCONE packets more frequently (see Section 8.1 of {{I-D.ietf-scone-protocol}}).
+
+If the CPU processing load in network elements for updating SCONE packets is not the
+limiting factor, it is preferred for endpoints to send, and network elements to update,
```
ihlar: This is pretty much right, but the endpoints can not take element CPU into consideration. 
Endpoints can send as frequently as they see fit, but the elements make decisions on update frequency based on considerations of CPU-load etc.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -129,10 +129,27 @@ network to signal the maximum allowable bit rate, and reduce CPU
...
+SCONE packets with a higher frequency than every 20 to 30 seconds to minimize the delay
+in activating the advised bit-rate due to packet loss. However, if the CPU processing
+load in a network element becomes a bottleneck, then the SCONE packet update frequency
+can be scaled back to 20 to 30 seconds. Essentially, there is a trade-off between the
+delay in activating the advised bit-rate due to packet loss versus the CPU processing load.
```
ihlar: There is an additional consideration here. 
A network that has dynamic policies will likely care much more about timely updates than a network that has fixed policies (e.g., subscription-based policies that are not intended to change over SCONE-like timescales).

draft-ietf-scone-applicability-manageability.md
```diff
@@ -129,10 +129,27 @@ network to signal the maximum allowable bit rate, and reduce CPU
...
+{{I-D.ietf-scone-protocol}}). Instead, the reliable delivery of throughput advice
+relies on the periodic sending of SCONE packets by the application endpoints.
+
+To ensure advice is not lost and does not expire, application endpoints are expected
+to send SCONE packets periodically, typically at least twice per monitoring period or
+every 20 to 30 seconds. The SCONE Network Element then takes these opportunities to update
+the throughput advice in the traversing packets. While endpoints could utilize QUIC
+transport-layer acknowledgments to detect lost datagrams and trigger a retransmission
+of a SCONE packet, it is generally simpler and highly effective to achieve reliability
+by simply sending SCONE packets more frequently (see Section 8.1 of {{I-D.ietf-scone-protocol}}).
```
mirjak: This paragraph should be removed as it is redundant to the spec and it's not good to say the same thing twice with different words.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -129,10 +129,27 @@ network to signal the maximum allowable bit rate, and reduce CPU
...
+
+If the CPU processing load in network elements for updating SCONE packets is not the
+limiting factor, it is preferred for endpoints to send, and network elements to update,
+SCONE packets with a higher frequency than every 20 to 30 seconds to minimize the delay
```
mirjak: Where does this number come from? This should not provide different/contradicting advice to the main spec.

draft-ietf-scone-applicability-manageability.md
smishra1200: Agree that Applicability document should guide operators rather than rehash core protocol mechanics. The above two paragraphs can be removed entirely so we are not repeating the endpoint sending behaviors already defined in Sections 7.1 and 8.1 of the main spec.

draft-ietf-scone-applicability-manageability.md
smishra1200: You are right to point this out and I agree. Endpoints have no visibility into a Network Element's CPU load and will send packets at their own pace. i can rewrite this paragraph to clarify that balancing CPU load and update frequency is strictly an operational configuration choice made by the network element, independently of the endpoint.

draft-ietf-scone-applicability-manageability.md
smishra1200: The 20 to 30 second value was actually drawn directly from Sections 7.1 and 8.1 of the core SCONE protocol specification, which state that updating every 20 to 30 seconds is likely sufficient. However, because per your above comments, I will be removing the redundant protocol explanations including removing these numbers entirely to avoid any perceived contradiction with the main spec.

draft-ietf-scone-applicability-manageability.md
smishra1200: This is an excellent operational consideration and exactly the kind of deployment guidance this document needs. I can incorporate this distinction into the text to highlight that dynamic policies require timely updates, whereas fixed subscription policies allow operators to scale back the update frequency to save CPU resources.

britram:
reviewed at the first appman interim, this one looks good to merge.

mirjak:
I think discussion is still on-going here, no?

smishra1200:
@mirjak, Updated text and heading.  Please see if this addresses issues?


Major updates include:
1. Removed the phrase "the network element must make independent operational decisions". Softened it to "needs to make independent operational decisions" to ensure we aren't creating new pseudo-requirements outside of the core protocol document.
2. Changed the title from "Retransmission of Rate of Advised Bit-rates" to "Mitigating Packet Loss".

Please see the revised text:

**Mitigating Packet Loss:**
Packet loss or non-delivery of SCONE advice directly reduces its effectiveness. Because the reliable delivery of throughput advice relies entirely on the periodic sending of SCONE packets by application endpoints, the network element needs to make independent operational decisions on how frequently to update those traversing packets. This decision relies on operational considerations such as CPU load and the nature of the network policies. 

A network enforcing dynamic policies might prioritize updating SCONE packets immediately upon a policy trigger to minimize the application's reaction time to the new limit. Conversely, a network enforcing fixed, subscription-based policies can safely scale back its update frequency to conserve CPU resources, provided it still updates SCONE packets periodically (e.g., every 20 to 30 seconds). This periodic update frequency ensures that the throughput advice reliably reaches the endpoint and does not inadvertently expire across the standard monitoring period due to normal packet loss.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -128,11 +128,19 @@ UDP 4-tuple. Such hints prevent unnecessary default rate-limiting, allow the
...
+sending of SCONE packets by application endpoints, the network element needs to make
+independent operational decisions on how frequently to update those traversing packets.
+This decision relies on operational considerations such as CPU load and the nature of
+the network policies. A network enforcing dynamic policies might prioritize updating SCONE
+packets immediately upon a policy trigger to minimize the application's reaction time to the
+new limit. Conversely, a network enforcing fixed, subscription-based policies can safely
+scale back its update frequency to conserve CPU resources, provided it still updates SCONE
+packets periodically (e.g., every 20 to 30 seconds). This periodic update frequency ensures
+that the throughput advice reliably reaches the endpoint and does not inadvertently expire
+across the standard monitoring period due to normal packet loss.
```
huitema: Could we add a reference to the Scone Transport spec, section 7.1.  Applying Throughput Advice Signals?

I think that what we are addressing here are an area voluntarily left unspecified by the specification, because of the tension between reliability and CPU load. The transport specification acknowledges that the network elements might not process every SCONE packet, and only mandate that "Senders that send a SCONE packet or network elements that update SCONE packets every 20–30 seconds is likely sufficient to ensure that throughput advice is not lost."

A conforming network element will process at least one packet every "20 to 30 second", but may want to process them more often, either to better mitigate packet losses or to ensure timely notifications. The appman document could say that and provide guidance.

Also, the title of this section should not be just about packet loss. It is about reliability, load and update frequency.

smishra1200: Hi Christian, Yes, I can definitely add the reference to the  Section 7.1 "Applying Throughput Advice Signals" and retune the text. 

RE the title change, I will note that we have a separate sub-section on Frequency of Updates and also a sub-section on Considerations of Processing Complexity that discusses load, so want to avoid, using frequency & load in the title, but, how about " Reliability & Mitigating Packet Loss"? 

I can update with a new PR with above changes including some tweaking of the text? Sound good?

smishra1200: @huitema, here is the proposed text and I can submit this on top of the last PR subject to feedback. Here is the text:

**3.4. Reliability and Mitigating Packet Loss**
Packet loss or non-delivery of SCONE advice directly reduces its effectiveness. Because the reliable delivery of throughput advice relies entirely on the periodic sending of SCONE packets by application endpoints, Section 7.1 ("Applying Throughput Advice Signals") of {{I-D.ietf-scone-protocol}} leaves the exact update frequency flexible. This flexibility allows operators to manage the tension between signaling reliability and network CPU load.

Ideally, a network element updates advice in SCONE packets at least twice per the 67-second monitoring period (approximately every 20 to 30 seconds), but operators may choose to process and update SCONE packets more frequently to better mitigate packet losses or to ensure timely notifications to the application. Therefore, the network element needs to make independent operational decisions on how frequently to update those traversing packets. A network enforcing dynamic policies might prioritize updating SCONE packets immediately upon a policy trigger to minimize the application's reaction time to the new limit. Conversely, a network enforcing fixed, subscription-based policies can safely scale back its update frequency to the 20 to 30-second baseline to conserve CPU resources. This baseline periodic update frequency ensures that the throughput advice reliably reaches the endpoint and does not inadvertently expire across the standard 67-second monitoring period due to normal packet loss.

Operators balancing this reliability against network element overhead can refer to the sections on Frequency of Updates and Considerations of Processing Complexity for further guidance.

huitema: Looks good, except for the word "Ideally". This is not an abstract ideal, it is the spec. Saying something like "A network element updates advice in SCONE packets at least twice per the 67-second monitoring period.." would be less ambiguous. Also, this is twice for each 4-tuple, but I guess this is discussed elsewhere.

smishra1200: @huitema,  I have removed "Ideally" to make the baseline expectation unambiguous. The sentence now says, "A SCONE-enabled network element updates advice in SCONE packets at least twice per the 67-second monitoring period (approximately every 20 to 30 seconds),....". And, yes,  the 4-tuple is covered explicitly in Flow Awareness and Per-Flow Signaling, which notes that throughput advice is applied on a UDP 4-tuple basis, so we don't need to repeat it in this section. 

I have submitted the revised PR.

huitema (approved):
Looks good. Thanks.

Issue #24, opened 2026-04-14
Retransmission vs send upated advice more often every monitoring interval


billwuqin:

zaheduzzaman:
Can you please describe the issue better?

billwuqin:
I think the key difference between retransmission and periodical update is whether the the packet loss can be detected. For retransmission, I think it should be clear which layer loss detection mechanism is used? QUIC layer, or other layer, In addition, in case of multi-data stream within a QUIC connection, one stream loss can not affect another stream, should this be documented?

smishra1200:
> I think the key difference between retransmission and periodical update is whether the the packet loss can be detected. For retransmission, I think it should be clear which layer loss detection mechanism is used? QUIC layer, or other layer, In addition, in case of multi-data stream within a QUIC connection, one stream loss can not affect another stream, should this be documented?

This is covered in PR #33 which states, When evaluating compliance, network operators will need to account for the time required for SCONE packets to be updated, received by endpoints, and acted upon by the application. Operators can accommodate this by utilizing a sliding window approach. Specifically, as established in {{Dynamic Updates}} section, operators should evaluate QUIC flows against the highest throughput limit advised over the preceding two monitoring periods (a span of 134 seconds). If a network element cannot update the throughput advice in every traversing SCONE packet, operators might configure a longer sliding window to account for the possibility of packet loss.
To simplify the measurement function, reduce computational load, or offload this function to another node in the network, operators can select any value larger than the baseline 67-second window for their measurement and averaging period.

PR #31, opened 2026-04-18
Update "Dynamic Updates" Section (Issue #24)

smishra1200:
GitHub #24 . Rework Dynamic Update section to harmonize with SCONE protocol and provide clarity in the text.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -143,16 +143,19 @@ adjustable update intervals based on application requirements, network
 capacity, and operational constraints.
 
 ## Dynamic Updates
-Target throughput advice can change dynamically due to access network
-handovers or changes in subscriber policy (e.g., reaching data thresholds).
-Because a Network Element cannot synthesize or inject a new SCONE packet into
-a flow, it needs to wait for the next endpoint generated SCONE packet to traverse
-the path. Following a dynamic trigger, the Network Element can opportunistically
-update the very next available SCONE packet(s) rather than waiting for its normal
-interval, thereby minimizing the application's reaction time to the new network
-state (see Section 7.1 and Section 9.2 of {{I-D.ietf-scone-protocol}}).
-Details on trade-off between the delay in dynamic updates to the advised
-bit-rate versus the CPU processing load is also covered in Section {{freq-updates}}.
+The frequency of SCONE signaling is fundamentally driven by the application endpoint
+based on its specific traffic characteristics. Highly adaptive flows, such as ABR video
+clients fetching short media segments, may choose to send SCONE packets frequently
+(for example, matching their segment request cadence) to rapidly track network capacity
+changes. Conversely, non-ABR applications, for example bulk transfer applications such
+as background software updates, may function effectively by sending SCONE packets less
+frequently, as they do not require immediate reactions to changing network conditions.
+
+For network operators, understanding this application-driven behavior is important because
+the Network Element relies entirely on these endpoint-generated packets to communicate
+throughput advice. While the core SCONE protocol defines the baseline timers to prevent
+advice from expiring, operators should expect the actual frequency of passing SCONE packets
+to vary significantly depending on the application type.
 
 ## Monitoring and Logging
 SCONE signaling can be integrated into existing operational and
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -134,7 +134,7 @@ SCONE Network Elements and application endpoints should support retransmission o
 periodic re-sending of SCONE packets to ensure reliable delivery.
 Conformance depends on the behavior of both network and application endpoint.
 
-## Frequency of Updates
+## Frequency of Updates {#freq-updates}
 The rate at which SCONE updates are issued depends on flow
 characteristics and available computational resources. Excessively
 frequent updates may increase CPU load, while infrequent updates may
@@ -151,9 +151,8 @@ the path. Following a dynamic trigger, the Network Element can opportunistically
 update the very next available SCONE packet(s) rather than waiting for its normal
 interval, thereby minimizing the application's reaction time to the new network
 state (see Section 7.1 and Section 9.2 of {{I-D.ietf-scone-protocol}}).
-Additional details on trade-off between the delay in dynamic updates to the advised
-bit-rate versus the CPU processing load is also covered in details
-in Section {{Frequency of Updates}}.
+Details on trade-off between the delay in dynamic updates to the advised
+bit-rate versus the CPU processing load is also covered in {{Section freq-updates}}.
 
 ## Monitoring and Logging
 SCONE signaling can be integrated into existing operational and
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -143,15 +143,17 @@ adjustable update intervals based on application requirements, network
 capacity, and operational constraints.
 
 ## Dynamic Updates
-Dynamic rate limits updates can be enforced by the network during active
-application sessions due to:
-
-- Changes in access network type (requiring updated throughput advice)
-- Changes in Subscriber policy (e.g., exceeding usage thresholds)
-
-In such cases, the SCONE Network Elements need to be able to initiate SCONE
-packets to provide updated advice, or applications should generate SCONE
-packets frequently enough to trigger network responses.
+Target throughput advice can change dynamically due to access network
+handovers or changes in subscriber policy (e.g., reaching data thresholds).
+Because a Network Element cannot synthesize or inject a new SCONE packet into
+a flow, it needs to wait for the next endpoint generated SCONE packet to traverse
+the path. Following a dynamic trigger, the Network Element can opportunistically
+update the very next available SCONE packet(s) rather than waiting for its normal
+interval, thereby minimizing the application's reaction time to the new network
+state (see Section 7.1 and Section 9.2 of {{I-D.ietf-scone-protocol}}).
+Additional details on trade-off between the delay in dynamic updates to the advised
+bit-rate versus the CPU processing load is also covered in details
+in Section {{Frequency of Updates}}.
 
 ## Monitoring and Logging
 SCONE signaling can be integrated into existing operational and
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -152,7 +152,7 @@ update the very next available SCONE packet(s) rather than waiting for its norma
 interval, thereby minimizing the application's reaction time to the new network
 state (see Section 7.1 and Section 9.2 of {{I-D.ietf-scone-protocol}}).
 Details on trade-off between the delay in dynamic updates to the advised
-bit-rate versus the CPU processing load is also covered in {{Section freq-updates}}.
+bit-rate versus the CPU processing load is also covered in Section {{freq-updates}}.
 
 ## Monitoring and Logging
 SCONE signaling can be integrated into existing operational and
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -143,15 +143,16 @@ adjustable update intervals based on application requirements, network
...
-In such cases, the SCONE Network Elements need to be able to initiate SCONE
-packets to provide updated advice, or applications should generate SCONE
-packets frequently enough to trigger network responses.
+Target throughput advice can change dynamically due to access network
+handovers or changes in subscriber policy (e.g., reaching data thresholds).
+Because a Network Element cannot synthesize or inject a new SCONE packet into
+a flow, it needs to wait for the next endpoint generated SCONE packet to traverse
+the path. Following a dynamic trigger, the Network Element can opportunistically
+update the very next available SCONE packet(s) rather than waiting for its normal
+interval, thereby minimizing the application's reaction time to the new network
```
mirjak: I don't understand what is meant by "opportunistically update the very next available SCONE packet(s) rather than waiting for its normal interval". Can you explain further?

smishra1200: Section [9.2 ](https://datatracker.ietf.org/doc/html/draft-ietf-scone-protocol-04#name-damage-to-other-protocols)of SCONE Protocol explicitly states: "Network elements MAY update more often immediately after a change in their throughput advice, to reduce the reaction time from senders." 

Building on this, the PR uses the phrase "opportunistically update" to clarify that following a dynamic network change, a Network Element does not need to wait for the full 67-second monitoring period to elapse. Instead, it takes immediate advantage of the very next available endpoint-generated packet to deliver the updated throughput advice, minimizing the application's reaction time.

mirjak: of course it doesn't. Why would it not add a new value to the available scone packet and then, of course, re-start the monitoring at the same time or actually wait one monitoring period from that point on. I find this text confusing...

smishra1200: The PR is trying to say, if there are events such as change in network or subscription policy, network does not have to wait for the full 67-second window to update ( ie add a new value in SCONE packet). Network element can "time" such updates to coincide with the "change" in policy (whatever that may be). RE re-starting monitoring, I assume you mean, reset the 67-second timer. I think, this is a very good suggestion. Happy to add that text, but, if we want this as a requirement, then protocol document might be a better place? But, going back to the core of your question, what part is confusing you? Happy to rephrase the part which is not clear or feel free to suggest what you would like to see?

mirjak: The point is I've never would have gotten the idea to wait with the update for 67s. Of course the update should be signalled as soon as possible! Where did you get the impression that there is any requirement to wait? It's only the enforce that needs to be paused after an update but that even more though underlines the point that an update should always be sent as fast as possible. Adding "opportunistically" is therefore re confusing because it seems to give people the idea that this is not the expected behavior.

smishra1200: The protocol explicitly states in Section 7.1 that to prevent throughput advice from expiring, a network element "needs to ensure that it updates throughput advice in SCONE packets with no more than a monitoring period [67 seconds] between each update". It goes on to recommend that ideally, they update advice at least twice per monitoring period.

Section 9.2 of the protocol document also establishes the allowance for dynamic changes, stating, "Network elements MAY update more often immediately after a change in their throughput advice, to reduce the reaction time from senders".

The intent of this PR text is to bridge these requirements to account for dynamic conditions and clarify that when a dynamic condition occurs, the NE applies the change immediately. (Presumably, the NE will also reset its 67-second keep-alive window at this point). I did not mean to imply there was a scheduled 67-second timer delay before sending the signal. 

I can reword this to clarify that the network element simply relies on the arrival of the next endpoint-generated packet to send the update immediately, and also remove the word "opportunistically" since it makes the expected fast signaling sound like an exception.

Could you also clarify your point that "only the enforce needs to be paused"? Section 7.2 states that a network element should base its monitoring on the maximum value from the preceding two monitoring periods to allow applications time to adjust. By pausing enforcement, do you mean the network element should start a fresh 67-second window after the latest update is sent before enforcing the new limit, or do you mean something else in terms of how that grace period is handled?

mirjak: There is no need to "bridge these requirements" because they are different requirements and completely clearly expressed in the protocol spec. The scenarios is the following:

And endpoint has to sent a scone packet at least every 67sec but best more often. Let's assume it sends one every 2s. Writing the current advice value into every scone packet every 2 second might be too much effort for an network element, so it might define it's own update interval, say every 10s,  and only write the current scone value into the next scone packet it sees after that interval expired. As long as it ends up sending the value at least once every 67s, this is sufficient to work wrongly because the signal is always valid for 67s even if a scone packet with a high values is received in-between. And yes, if the values changes it is recommended to not wait for 10s and just update the next scone packet that comes by.

However, this sufficient described in the scone protocol spec and there is no need to re-hash it in this document. If there anything unclear or even missing in the protocol spec, we should clarify it there!

mirjak: > Could you also clarify your point that "only the enforce needs to be paused"? Section 7.2 states that a network element should base its monitoring on the maximum value from the preceding two monitoring periods to allow applications time to adjust. By pausing enforcement, do you mean the network element should start a fresh 67-second window after the latest update is sent before enforcing the new limit, or do you mean something else in terms of how that grace period is handled?

What I meant is covered by these "two monitoring periods": you need one period to be sure that the signal is conveyed and then you start you first actual monitoring period but you can start enforcement only at the end of that monitoring period.

smishra1200: > There is no need to "bridge these requirements" because they are different requirements and completely clearly expressed in the protocol spec

I agree that the core mechanics are clearly expressed in the protocol spec and  to not rehash them. However, as discussed during IETF 125, the audience for the AppMan document is network operators, and its purpose is to serve as a deployment guide. While the protocol spec defines the boundaries, operators need explicit guidance on how to configure their Network Elements within those boundaries.

>  so it might define it's own update interval, say every 10s, and only write the current scone value into the next scone packet it sees after that interval expired. As long as it ends up sending the value at least once every 67s, this is sufficient to work wrongly because the signal is always valid for 67s even if a scone packet with a high values is received in-between. And yes, if the values changes it is recommended to not wait for 10s and just update the next scone packet that comes by.

Your example of configuring a 10-second (or whatever value a network operator deems suitable) update interval to optimize CPU load, and your exact breakdown of how to configure traffic policers for the two-period enforcement delay, are perfect examples of operational deployment strategies that belong in this document.

I am open to update and rewrite the section to remove any rehashing of the protocol rules. Instead, the text can strictly focus on operator configuration, using your examples for update intervals and enforcement tuning.

smishra1200: @mirjak wrote:

> What I meant is covered by these "two monitoring periods": you need one period to be sure that the signal is conveyed and then you start you first actual monitoring period but you can start enforcement only at the end of that monitoring period.

Your breakdown of using the first period to convey the signal and starting enforcement only at the end of the second period makes perfect sense, and that is exactly the operational guidance text that can be added.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -143,15 +143,13 @@ adjustable update intervals based on application requirements, network
...
-In such cases, the SCONE Network Elements need to be able to initiate SCONE
-packets to provide updated advice, or applications should generate SCONE
-packets frequently enough to trigger network responses.
+Target throughput advice can change dynamically due to access network handovers or changes in
+subscriber policy (e.g., reaching data thresholds). When such a change occurs, the network
+element can immediately update the next available traversing SCONE packet with the new
+throughput limit rather than waiting for its normal update interval, thereby minimizing the
+application's reaction time to the new network state. Additional details on the operational
+trade-offs between the delay in dynamic updates to the advised bit-rate versus the CPU
+processing load are covered in Section {{freq-updates}}.
```
mirjak: Do we still need this section?

britram:
reviewed at the first appman interim, this one needs additional discussion

draft-ietf-scone-applicability-manageability.md
smishra1200: @mirjak the reason I would say to consider keeping Dynamic Updates is because it is like a "Trigger" for actioning on immediate reactions. 

It tells the operator what to do when a sudden network event happens (like a handover or hitting a data cap). It advises them to immediately update the very next available packet, rather than waiting for the normal retransmission interval we discussed in the Retransmission section.

As far as the text is concerned, I have whittled it down to something like:

"Target throughput advice can change dynamically due to access network handovers or changes in subscriber policy (e.g., reaching data thresholds). When such a change occurs, the network element can immediately update the next available traversing SCONE packet with the new throughput limit rather than waiting for its normal update interval, thereby minimizing the application's reaction time to the new network state."

I'm also open to re-evaluate once we have the full document. I am beginning to lose track of what we have and where. I do suspect, we prolly are saying the same thing at two different places.

gorryfair: I like the idea behind writing something about dynamic changes to the guidance, but I'm unsure what /rather than waiting for its normal update interval/ ...refers to?

gorryfair: Also be careful on what the "guidance" is formally called!

smishra1200: @gorryfair, Good catch. The phrase "normal update interval" was ambiguous since the protocol doesn't define a rigid network-side timer. We were referring to the periodic update cadence (e.g., 20-30 seconds) as now stated in the "Mitigating Packet Loss" section to ensure the signal doesn't expire. I have submitted an updated the text to explicitly reference that periodic update cadence so it's clear what the network element is waiting for. 

Also changed "throughput limit" in that sentence to "throughput advice". Here is the text for Dynamic Updates:
Target throughput advice can change dynamically due to access network handovers or changes in subscriber policy (e.g., reaching data thresholds). When such a change occurs, the network element can immediately update the next available traversing SCONE packet with the new throughput advice, rather than waiting for its next periodic update (such as the cadence established to mitigate packet loss). This immediate update minimizes the application's reaction time to the new network state.

billwuqin:
How this section is related to section 5.4 Following Throughput Advice and section 6.3 on Indications for Migrated Flows in SCONE base protocol? 
Be careful the boundary with SCONE protocol.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -161,15 +161,12 @@ traversing packets at highly variable, application-driven intervals rather than
...
-
-- Changes in access network type (requiring updated throughput advice)
-- Changes in Subscriber policy (e.g., exceeding usage thresholds)
-
-In such cases, the SCONE Network Elements need to be able to initiate SCONE
-packets to provide updated advice, or applications should generate SCONE
-packets frequently enough to trigger network responses.
+Target throughput advice can change dynamically while a flow is active, for example when a
+subscriber reaches a data threshold or a network policy changes. When this happens, the network
+element can immediately update the next traversing SCONE packet with the new throughput advice,
```
martinthomson, line 166:
```suggestion
element can update the next traversing SCONE packet with the new throughput advice,
```

"immediately" is not possible.  You might consider saying more about this, because "minimizing" is doing a lot of work in that final sentence.

martinthomson (commented):
This is better than the original text, which used "need to" in a way that was somewhat futile.  However, the use of "immediate" here is still a bit misleading.

draft-ietf-scone-applicability-manageability.md
smishra1200: Thanks, Martin. Agreed on both. I dropped "immediately" since a network element has to wait for the next endpoint-generated SCONE packet, and I reworked the final sentence so "minimizing" is not carrying the weight. It now says how soon the application sees the change depends on when the next SCONE packet traverses, since the network element cannot originate one. The revised text is in PR #46, where this is now being tracked under issue #24.

Issue #25, opened 2026-04-14
How do you know SCONE signaling fails to reach the intended  endpoints?


billwuqin:
Packet loss detection, set timeout? Feedback from the intended endpoints?

gorryfair:
P.S. The answer could fall into two parts: How does the sender detect? How can a network operator understand?

smishra1200:
Also, please refer to the following text in "Retransmission Rate of Advised Bit-Rate:
[To ensure advice is not lost and does not expire, application endpoints are expected
to send SCONE packets periodically, typically at least twice per monitoring period or
every 20 to 30 seconds. The SCONE Network Element then takes these opportunities to update
the throughput advice in the traversing packets. While endpoints could utilize QUIC
transport-layer acknowledgments to detect lost datagrams and trigger a retransmission
of a SCONE packet, it is generally simpler and highly effective to achieve reliability
by simply sending SCONE packets more frequently](url)

Issue #7, opened 2026-03-22
In Section 1: Forwarding Changes & Network Elements


gorryfair:
I expect many operators will need to think through is that they do not have sole right to change the SCONE information, rather any network elememt along the path could reduce the value. That's not a bad thing, and if we work out how to do SCONE with WiFi wisely, it could even be common on some paths. Is this discussed somewhere? I therefore expected to see a short paragraph explaining how the SCONE endpoint would operate with no SCONE network element on path and also with multiple network elements on the same path.

- I expected a para on what happens when a forwarding path change results in a change of the SCONE Network Element that provides the advice. Section 6.1 of the protocol Spec suggests that missing the start of a flow (because a traffic starts on a different path, could result in the SCONE Network element on the new path not seeing the start of the flow. I'd hope this case would be explained and the result highlighted as an example.

smishra1200:
Gorry wrote:

> I expect many operators will need to think through is that they do not have sole right to change the SCONE information, rather any network elememt along the path could reduce the value. That's not a bad thing, and if we work out how to do SCONE with WiFi wisely, it could even be common on some paths. Is this discussed somewhere?

From the operator's perspective, Section 7.1 of the SCONE Protocol guarantees that multiple network elements can coexist safely.
From the application's perspective, we cover this in our newly created section 3.7: "The application endpoint handles the Rate Signal field present in the SCONE packet as specified in the SCONE Protocol, irrespective of how many network elements are present. For example, if an endpoint receives multiple SCONE packets within a 67-second monitoring period with different Rate Signal values, it is expected to apply the lowest throughput advice received (see Section 5.4 of {{I-D.ietf-scone-protocol}})."

Do this adequately addresses your question?

smishra1200:
> I expected a para on what happens when a forwarding path change results in a change of the SCONE Network Element that provides the advice. Section 6.1 of the protocol Spec suggests that missing the start of a flow (because a traffic starts on a different path, could result in the SCONE Network element on the new path not seeing the start of the flow. I'd hope this case would be explained and the result highlighted as an example.
"We address this scenario in the newly revised Section 3.7 (Dynamic Updates) of the AppMan draft, which explicitly covers "access network handovers" and explains how the new Network Element reacts by immediately updating the next available packet."

smishra1200:
Closing this issue, please see PR #32

PR #32, opened 2026-04-18
Update Introduction & add two new sections

smishra1200:
GitHub #7 and #25. Address issue raised on no network element on path, single network element or multiple network element on path and the expected behavior. Introduced topic in the Introduction section and added two new sub sections "Presence of SCONE NE on the Data Path" and "Change of Network Element During an Active Flow" to account for connection migration.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -82,15 +82,28 @@ mechanisms,to proactively align their transmission rates with network policies. 
 Applicability and Manageability considerations for deploying the SCONE protocol within service provider networks.
 It also addresses operational, configuration, and management aspects not covered in the core protocol specification.
 
-To participate in SCONE, a network element is assumed to have the functional capability to identify and track scone compliant application flows, recognize and process SCONE packets within those flows and map network policies into
-throughput advice to be inserted into the SCONE packets. While on-path network elements may exist at various points
-between the server and the client application end-points, their specific configuration and role will influence the
-advice they generate. Different network architectures handle flow visibility and policy enforcement at different points.
-In mobile networks, for example, the User Plane Function (UPF) in 5G {{5G-Arch}} and the Packet Data Network Gateway (P-GW) in 4G network {{4G-Arch}} can generate throughput advice to guide ABR applications on a per-flow basis. In contrast,
-other environments, such as wireline broadband or Wi-Fi, may apply policies at centralized aggregation points or gateways such as the Broadband Network Gateway serving multiple devices.
+To participate in SCONE, a network element is assumed to have the
+functional capability to identify and track SCONE-compliant QUIC
+flows, recognize and process SCONE packets within those flows, and
+map network policies into throughput advice to be inserted into the
+SCONE packets. However, from an end-to-end perspective, a data path
+might contain no such network elements, a single network element, or
+multiple network elements. When multiple elements are present, they
+can update the throughput advice independently without any coordination.
+More detail is covered in subsequent sections.
 
-Encompassing deployment of network elements in a wide range of networks, this document is limited to discussing the core Applicability and Manageability considerations for the SCONE protocol to ensure its consistent and effective use across
-varied network paths.
+When on-path network elements are present between the server and the client
+application end-points, their specific configuration and role will influence the advice they
+generate. Different network architectures handle flow visibility and policy enforcement at
+different points. In mobile networks, for example, the User Plane Function (UPF) in 5G {{5G-Arch}}
+and the Packet Data Network Gateway (P-GW) in 4G network {{4G-Arch}} can generate throughput
+advice to guide ABR applications on a per-flow basis. In contrast, other environments,
+such as wireline broadband or Wi-Fi, may apply policies at centralized aggregation points
+or gateways such as the Broadband Network Gateway serving multiple devices.
+
+Encompassing deployment of network elements in a wide range of networks, this document
+is limited to discussing the core Applicability and Manageability considerations for
+the SCONE protocol to ensure its consistent and effective use across varied network paths.
 
 
 # Terms and Definitions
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -153,6 +166,65 @@ In such cases, the SCONE Network Elements need to be able to initiate SCONE
...
+A change in the on-path network element can occur when an application changes
+its access network. This typically manifests in two scenarios: QUIC connection
+migration where the IP address changes, or mobility events where the IP address
+remains unchanged. In both cases, the endpoint and the new network element can
+follow the steps for migrated flows specified in Section 6.3 of
+{{I-D.ietf-scone-protocol}}. Because a connection that migrates cannot assume
+that throughput advice from one path applies to a new path, endpoints can send
+SCONE packets early on the new path. This gives the network element on the new
+path an opportunity to detect the SCONE-capable flow and provide its own throughput
+advice.
```
mirjak: This text seems very redundant with the SCONE spec.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -153,6 +166,65 @@ In such cases, the SCONE Network Elements need to be able to initiate SCONE
...
+- New network element does not support SCONE (or misses the flow): If the new
+network element does not support the SCONE protocol, or if it fails to detect
+the flow after migration, the endpoint will cease receiving SCONE updates.
+The endpoint handles this scenario exactly as it would any loss of signal.
+If a full 67-second monitoring period elapses without receiving throughput
+advice, the previous advice expires (see Section 5.4 of {{I-D.ietf-scone-protocol}}).
+The endpoint can then assume that either there are no SCONE-aware network elements
+on the new path, or that they do not wish to enforce a specific bit-rate, allowing
+the endpoint to remove any SCONE-advised constraints and operate without SCONE
+limits for that flow.
```
mirjak: I don't think it necessary to discussion this here as this about what an scone endpoint needs to do and hat is covered in the scone protocol spec.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -153,6 +166,65 @@ In such cases, the SCONE Network Elements need to be able to initiate SCONE
 packets to provide updated advice, or applications should generate SCONE
 packets frequently enough to trigger network responses.
 
+## Presence of SCONE Network Elements On the Data Path
```
mirjak, line 165: I don't know why this is needed in the document. This doesn't provide any advise to the network, nor to the application using scone. It only restated what is defined by the spec on how a client is required to handle the throughput advise. I don't support adding this section. In generally I think we should rather try to keep the document as short as possible and already remove some of the existing redundant and duplicated text regarding the protocol spec.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -82,15 +82,28 @@ mechanisms,to proactively align their transmission rates with network policies.
...
-Encompassing deployment of network elements in a wide range of networks, this document is limited to discussing the core Applicability and Manageability considerations for the SCONE protocol to ensure its consistent and effective use across
-varied network paths.
+To participate in SCONE, a network element is assumed to have the
+functional capability to identify and track SCONE-compliant QUIC
+flows, recognize and process SCONE packets within those flows, and
+map network policies into throughput advice to be inserted into the
+SCONE packets. However, from an end-to-end perspective, a data path
+might contain no such network elements, a single network element, or
+multiple network elements. When multiple elements are present, they
+can update the throughput advice independently without any coordination.
```
mirjak: note that it should only be updated to a lower value, even though there is of course no guarantee for that.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -82,15 +82,28 @@ mechanisms,to proactively align their transmission rates with network policies.
...
+More detail is covered in subsequent sections.
+
+When on-path network elements are present between the server and the client
+application end-points, their specific configuration and role will influence the advice they
+generate. Different network architectures handle flow visibility and policy enforcement at
+different points. In mobile networks, for example, the User Plane Function (UPF) in 5G {{5G-Arch}}
+and the Packet Data Network Gateway (P-GW) in 4G network {{4G-Arch}} can generate throughput
+advice to guide ABR applications on a per-flow basis. In contrast, other environments,
+such as wireline broadband or Wi-Fi, may apply policies at centralized aggregation points
+or gateways such as the Broadband Network Gateway serving multiple devices.
```
mirjak, line 98: Not sure why this paragraph is really important? I think that goes without saying? Why is are mobile network element highlighted specifically?

draft-ietf-scone-applicability-manageability.md
smishra1200: @mirjak. Agree, this is specifically covered in section "**Presence of SCONE Network Elements on the Data Path**", that says, "...For example, if an endpoint receives multiple SCONE packets within a 67-second monitoring period with different Rate Signal values, it is expected to apply the lowest throughput advice received...."

draft-ietf-scone-applicability-manageability.md
smishra1200: @mirjak 
> Not sure why this paragraph is really important? I think that goes without saying?

The introduction section was updated to answer question raised by Gorry in GitHub Issue #7  specifically, part of the comment ss "I therefore expected to see a short paragraph explaining how the SCONE endpoint would operate with no SCONE network element on path and also with multiple network elements on the same path."

> Why is are mobile network element highlighted specifically?
WiFI and wireline broadband are also included above, but when referring to mobile, the text is just highlighting both 4G and 5G and the likely place where similar function resides today.

draft-ietf-scone-applicability-manageability.md
smishra1200: Please see my response above and submitted an updated shortened version.

draft-ietf-scone-applicability-manageability.md
smishra1200: @mirjak,  I have pruned this section to removed redundant protocol text. Rewrote it to focus strictly on the operational takeaway for the network operator, for e.g., "... because SCONE is stateless, handovers require no explicit teardown or state-transfer between network elements." 

Submitting a PR along with the section Presence of SCONE Network Element on The Data Path".

Let me know if this leaner, operations-focused version looks better to you

britram:
reviewed at the first appman interim, I think this one is ready to merge, with questions about redundant or unnecessary text to be handled on the resulting document rev,

mirjak:
I'd prefer to conclude the discussion first and not merge new text that we would just remove later again; that's seems confusing.

draft-ietf-scone-applicability-manageability.md
smishra1200: @mirjak, I do agree on some points here. For one,  the text isrestating how the client handles the signal and repeating the protocol rules from Section 7.1 about replacing signals with lower values.

However, there is a core operational insight here that network planners need when designing their deployments, which is, that multiple network elements (if and when they exists) do not require control-plane coordination or synchronization.

And to address your questions, for one, I have cut down a lot of text. For e.g., I have  removed the zero/single element scenarios (since they offer no deployment advice) and deleted all the redundant protocol mechanics. 

The section is now just three sentences focused purely on the deployment architecture, assuring operators that they can deploy SCONE elements independently without building integration between them. Let me know if this reduced, operator-focused version resolves your concern?

Text:
**Presence of SCONE Network Elements on the Data Path**

In topologies where multiple SCONE-capable network elements exist on the same data path (for example, a mobile UPF and a downstream enterprise Wi-Fi gateway), operators can deploy and manage these elements completely independently. Operators do not need to implement control-plane coordination or synchronization between different network elements to negotiate rate limits. 

Because the core SCONE protocol inherently resolves multiple signals by preserving the most restrictive throughput advice along the path, each network element can simply apply its own local policy without requiring awareness of the presence, capabilities, or configuration of other network elements.

smishra1200: and also the text addresses question that Gorry raised to talk about zero network element, or one or more scone aware network element(s) and migration - Issue #7 . If above makes sense, I can do a PR with above text.

smishra1200:
Please see responses posted above

draft-ietf-scone-applicability-manageability.md
gorryfair: Could easily be fixed to "update (i.e. reduce)" ... and avoid this discussion?

smishra1200: @gorryfair could you please clarify which specific section & sentence you are asking to revise?

gorryfair: I was suggesting:
/can update the throughput advice independently/
can update (i.e. reduce) the throughput advice independently/

smishra1200: @gorryfair Got it. I updated the PR as suggested, ie s/can update the throughput advice independently/can update (i.e. reduce) the throughput advice independently/

billwuqin (commented):
The proposed change in the introduction is overlapping with the text proposed for #39, no?
please make sure to focus on generic aspect, make sure the text in the introduction aligns with what has been discussed in the main section 3.

billwuqin (approved):
Look better than the previous version, thanks for addressing my comment.

Issue #11, opened 2026-03-22
In Section 3.9: Please clarify words (NiT) around compliance


gorryfair:
Can this document explain what compliance means with respect to SCONE?

smishra1200:
> Can this document explain what compliance means with respect to SCONE?
Yes, expanded this section. for example added text to note, "Networks providing SCONE throughput advice can implement mechanisms to monitor QUIC flows and measure conformance to the advised bit-rate, either per flow of packets on the same UDP address tuple, or in aggregate across multiple QUIC flows if they contribute to a shared policy limit (e.g., at a device or network subscription level...." Please refer to PR #33.

PR #33, opened 2026-04-18
Update "Conformance Monitoring" Section

smishra1200:
GitHub #11 and #22. This section is a full rewrite of Conformance Monitoring and aligns with SCONE Protocol.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -142,7 +142,7 @@ reduce advisory effectiveness. Network Operators can define
 adjustable update intervals based on application requirements, network
 capacity, and operational constraints.
 
-## Dynamic Updates {#dynamic-updates}
+## Dynamic Updates
 Dynamic rate limits updates can be enforced by the network during active
 application sessions due to:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -163,12 +163,53 @@ isolation. Metrics of interest include:
...
+to simplify the measurement function, reduce computational load, or offload this function
+to another node in the network, they can select any value larger than 67 seconds as their
+measurement and averaging window.
+
+Some applications will not support SCONE, and others either will not or cannot follow the
+provided throughput advice. If the monitoring function indicates that an application is
+not respecting the signaled throughput advice, the network can use other rate-limiting
+policy enforcement mechanisms, such as dropping or delaying packets, to ensure the QUIC
+flow does not exceed the throughput limits set by network policy
+(see {{I-D.ietf-scone-protocol}}, Sections 7.2 and 7.3 ).
```
mirjak: or not do anything at all. We should note that the scone signal is also valuable to provide without monitoring and compliance enforcement!!!

smishra1200: Good point. Agree that the pure advisory nature of SCONE is highly valuable on its own, and it should not be implied that operators are expected to deploy strict rate limiters just because they are sending throughput advice. Happy to update the text to explicitly state that operators can choose to do nothing at all regarding enforcement, treating the SCONE signal as purely informational for cooperative endpoints.

britram:
reviewed at the first appman interim, this one looks good to merge, with potential follow up discussion on the resulting document revision.

Issue #23, opened 2026-04-14
SCONE at the application level or transport level?


billwuqin:
Appman said:
ECN/L4S based congestion control works at transport level and SCONE works at application level.

smishra1200:
> Appman said:
ECN/L4S based congestion control works at transport level and SCONE works at application level.
Agree. Have submitted PR #34 and there is more to come to clean up. We can review once the PR is added and see if it clarifies?

Issue #16, opened 2026-04-09
Add discussion about the relation of scone and L4S


mirjak:

mirjak:
Section 3.11 abut congestion control and also mentioned L4S but it would be good to talk more about when it makes sense to use one of the other or both. the protocol spec already explains that scone does not replace congestion control. the applicability statement should focus more on when to use things and how.

Issue #15, opened 2026-03-22
In Section 3.2: Please clarify position for ECN (NiT)


gorryfair:
Some words explaining that the SCONE marking for a flow is independent of the ECN Field may be useful - since ECN appeared many times in the discussion of SCONE.

- Ought this to include a reference to ECN?

Issue #12, opened 2026-03-22
In Section 3.11: Please clarify words (NiT) about "stable".


gorryfair:
Can you quantify relatively stable with an example, would  an example be greater than multiple seconds? 10s seconds? etc

smishra1200:
> Can you quantify relatively stable with an example, would an example be greater than multiple seconds? 10s seconds? etc
The whole section is entirely revised and have removed "relatively stable", the text says, "SCONE throughput advice operates over a much longer period. SCONE can be used for any use case that requires network operators to send the maximum allowable bit-rate (throughput advice) that is driven by operator video policy, subscriber policy, and load management policy operating over a longer time period." Please have a look at the revised text in PR #34

PR #34, opened 2026-04-19
Rework 'interworking with Other Congestion Control Mechanism

smishra1200:
Address multiple GitHub issues (#12, #15, #16, and #23). The entire section is rewritten to attempt to address issues. The section outlines SCONE behavior and is distinct with CC mechanisms such as ECN and L4S.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -65,6 +65,16 @@ informative:
       - name: 3GPP
     date: 2025-01-07
 
+  RFC3168:
+    target: https://www.rfc-editor.org/info/rfc3168
+    title: The Addition of Explicit Congestion Notification (ECN) to IP
+    author:
+      - name: K. Ramakrishnan
+      - name: S. Floyd
+      - name: D. Black
+    date: 2001-09
+    seriesinfo: "RFC 3168"
+
 --- abstract
 
 This document describes the Applicability and Manageability considerations for providing throughput guidance to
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -175,20 +185,43 @@ SCONE signaling is expected to traverse the existing data path associated
...
-works at application level. SCONE does not replace the need for endpoints to perform
-congestion control or network to provide explicit or implicit congestion signals; rather,
-it complements these mechanisms by providing a variable range for application-level rate
-adaptation. In environments where both are present, SCONE and congestion control mechanisms
-co-exist: congestion control manages the immediate dynamics of the bottleneck link, while
-SCONE informs the application of the maximum sustained rate allowed by policy. Network Operators
-would benefit from harmonizing multiple congestion signaling methods by policy or scope
-deployments to avoid conflicting feedback.
+SCONE throughput advice is not a substitute for congestion feedback or congestion control
+mechanisms, such as Explicit Congestion Notification (ECN) or Low Latency, Low Loss, and
```
gloinul: The sentence feels strange as it talks about congestion control mechanisms, where still packet loss and delay are the primary signals, and where ECN and L4S are additional signals. So maybe reformulate this sentence to: 

SCONE throughput advice is not a substitute for congestion control
mechanisms in transport protocols, utilizing congestion feedback and signals such as acknowledgements, Explicit Congestion Notification (ECN), and Low Latency, Low Loss, and Scalable Throughput (L4S) (see Section 3.1 of {{I-D.ietf-scone-protocol}}).

gloinul (commented):
I think the text is an improvement. I have a comment on the wording in the first paragraph.

Maybe add some text to clarify that for flows using L4S what is generally written here do apply, L4S improves the response to the ECN congestion signal through its proportional marking.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -175,20 +185,43 @@ SCONE signaling is expected to traverse the existing data path associated
...
+Scalable Throughput (L4S) (see Section 3.1 of {{I-D.ietf-scone-protocol}}). Rather, they
+are complementary. Congestion signals provide real-time information on loss, delay, and
+transient congestion for a network path, typically operating on the time scale of a
+round-trip time (RTT), whereas SCONE throughput advice operates over a much longer
+period. SCONE can be used for any use case that requires network operators to send the
```
mirjak: I think it would be more important to talk about application use cases here. The network doesn't really have any sue cases because it doesn't really know what the traffic is; it only knows that the traffic is scone-enabled or not and can either support scone or not. Of course there is the question on what values to set the throughput advise to but I wouldn't call that a use case; it's more depend on what kind of information you have available in the network element and as such the advise might be static or adapted more dynamically.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -175,20 +185,43 @@ SCONE signaling is expected to traverse the existing data path associated
...
+maximum allowable bit-rate (throughput advice) that is driven by operator video policy,
+subscriber policy, and load management policy operating over a longer time period. The
+specific reasons or methods by which an operator determines these enforcement policies
+are out of scope for this document.
```
mirjak: Why is that out of scope? I don't think we need to discuss those policies but that a topic which I think would be explicitly in scope for a manageability document.

mirjak: However, note that I don't think it should be covered in this section. There is already another section covering this roughly currently named "Dynamic updates".

draft-ietf-scone-applicability-manageability.md
```diff
@@ -175,20 +185,43 @@ SCONE signaling is expected to traverse the existing data path associated
...
+Moreover, congestion could be experienced at an entirely different point on the network
+path than the network element providing the SCONE signal. Therefore, endpoints need to always
+respect the send rate constraints established by their congestion controller, even if the
+SCONE throughput advice indicates a higher available rate
+(see Section 3.1 of {{I-D.ietf-scone-protocol}}).
+
+In environments where both are present, SCONE and congestion control mechanisms co-exist.
+Congestion control manages the immediate dynamics of the bottleneck link, while SCONE informs
+the application of the maximum rate allowed by network policy. Network operators would benefit
+from harmonizing multiple congestion signaling methods by policy or within scoped deployments
```
mirjak: I don't understand what you mean by "harmonising"? Don't you say above rightfully that these signal are independent?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -175,20 +185,43 @@ SCONE signaling is expected to traverse the existing data path associated
...
+
+Because a congestion controller needs to detect changed conditions and alter sending
+behavior more quickly than SCONE allows, congestion signals can indicate a throughput limit
+that is different from the signaled SCONE throughput advice. Endpoints cannot assume that the
+rate indicated in throughput advice is achievable if congestion signals indicate otherwise.
+Moreover, congestion could be experienced at an entirely different point on the network
+path than the network element providing the SCONE signal. Therefore, endpoints need to always
+respect the send rate constraints established by their congestion controller, even if the
+SCONE throughput advice indicates a higher available rate
+(see Section 3.1 of {{I-D.ietf-scone-protocol}}).
```
```suggestion
(see Section 3.1 of {{I-D.ietf-scone-protocol}}).
In other words, in the best case the throughput advice is below the congestion limit
and when the application adheres to the advice, congestion control would be
application-limited and not go into action. However, in cases of e.g. high load,
congestion control would limit the throughput below the provided advice
as the scone advice is only an upper limit. As such congestion control or
a similar mechanism to react to congestion, e.g. a circuit breaker,
is always needed, in addition to SCONE.
```

draft-ietf-scone-applicability-manageability.md
smishra1200: This is an excellent point and definitely needs the clarification. I have incorporated your suggested reformulation into the text to reflect how transport protocols utilize both primary feedback (acknowledgments) and additional explicit signals (ECN/L4S). Thank you for catching this.

draft-ietf-scone-applicability-manageability.md
smishra1200: You make a very fair point. Since this is a Manageability document, explicitly declaring methods or policy determination "out of scope" may create unnecessary confusion. Also, as you correctly noted, we already provide the necessary context for what drives these policies (such as access network changes or subscriber limits) in the "Dynamic Updates" section, so we don't need to rehash it here. I can simply remove the whole "out-of-scope" sentence from the congestion control section so it remains tightly focused on L4S and ECN interworking.

draft-ietf-scone-applicability-manageability.md
smishra1200: You are completely right. The network is fundamentally unaware of the specific application traffic and doesn't have use cases per se. It simply provides static or dynamically adapted advice based on its policies. I will  remove the phrase "use case" when referring to the network. I can add a clarification to this section explicitly stating that the network provides the policy-driven advice (whether static or dynamic), while it is up to the applications (such as ABR video or bulk transfers) to apply that advice to their specific use cases. 
So, the first paragraph (adding Magnus' suggestion and also removing the out of scope and use case):

"SCONE throughput advice is not a substitute for congestion control mechanisms in transport protocols utilizing congestion feedback and signals such as acknowledgments, Explicit Congestion Notification (ECN), and Low Latency, Low Loss, and Scalable Throughput (L4S). Rather, they are complementary. Congestion signals provide real-time information on loss, delay, and transient congestion for a network path, typically operating on the time scale of a round-trip time (RTT). In contrast, SCONE throughput advice operates over a much longer period. Because the network element is generally unaware of the specific application traffic, it simply provides static or dynamically adapted advice based on available policy information. Operators can use SCONE to communicate these maximum allowable bit-rates driven by video optimization, subscriber data, or load management policies, independent of instantaneous link congestion. It is then up to the applications, such as adaptive bitrate video clients or bulk transfer tasks, to utilize this advice according to their specific use cases."

draft-ietf-scone-applicability-manageability.md
smishra1200: Thank you for the excellent suggested text! I have accepted your phrasing directly here in the PR to incorporate it into the draft.

Just a quick note: because we are making a few additional updates to this section (to address Magnus's feedback on acknowledgments and your earlier comments regarding L4S co-deployment and network use cases), I will be submitting a broader revised text for this section a bit later. That upcoming revision will fully integrate and preserve this phrasing you've provided. Thanks again for the help!

draft-ietf-scone-applicability-manageability.md
smishra1200: @mirjak. Valid point. Our intent was purely operational and advisory. We meant that network operators should ensure their policy configurations for SCONE limits and congestion control are consistent, so they don't give the application confusing or conflicting constraints. But, agree the current wording is just not helping to convey that. Here is the revised text. I am submitting a an update here with the whole section trying to incorporate all the above comments from you and @gloinul.

"In environments where both are present, congestion control manages the immediate dynamics of the bottleneck link, while SCONE informs the application of the maximum rate allowed by network policy. Network operators will benefit from ensuring that throughput advice policies and congestion control configurations are consistent within scoped deployments, to avoid providing conflicting feedback to applications."

Let me know if this get any better or worse :)

britram:
reviewed at the first appman interim, this one looks good to merge, with subsequent discussion on the resulting doc.

PR #36, opened 2026-04-19
Expand "Standards Compliance" Section

smishra1200:
GitHub #22. Expanded Standards Compliance Section. For example, added text including SCONE allows network operators and QUIC endpoints to exchange bit-rate information without custom APIs or per-network integrations and provide bit-rate guidance to self-adapt instead of relying on network rate limiters such as policers or shapers.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -171,8 +171,25 @@ implemented in a SCONE Network Element but may also be implemented
 elsewhere in the network.
 
 ## Standards Compliance
-SCONE signaling is expected to traverse the existing data path associated
-with the UDP 4-tuple flow for which the Network Element intends to send the advisory bit-rate.
+Because SCONE packets are always coalesced with ordinary QUIC packets, SCONE signaling
+operates entirely in-band. It does not introduce any additional routing overhead or
+require the creation of out-of-band signaling interfaces. Instead, SCONE signaling
+inherently traverses the already established network path, such as the existing
+connection between a user device and a network gateway, associated with the QUIC flow
+for which the network element intends to send throughput advice. This ensures that
+SCONE seamlessly integrates into existing architectures without requiring new tunnels
+or data paths to be established.
+
+By providing a standardized and scalable mechanism, SCONE allows network operators
+and QUIC endpoints to exchange bit-rate information without custom APIs or per-network
+integrations. SCONE improves user experience by enabling the network to provide bit-rate
+guidance directly to applications, allowing them to self-adapt instead of relying on
+network rate limiters such as policers or shapers. This avoids packet drops and throttling,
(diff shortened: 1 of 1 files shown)


ihlar:
This text reads a bit like marketing, which isn't necessarily a bad thing. However, I think that text like this might be better suited in an introduction section.

smishra1200:
@ihlar wrote:
"Standards Compliance" Section #36 

> This text reads a bit like marketing, which isn't necessarily a bad thing. However, I think that text like this might be better suited in an introduction section.

You are correct that it sounds like a trying to sell SCONE to the operators and in a way, we added this text to pick up a comment on the MIC, something like, should I implement and deploy this protocol or not?, so we were trying to answer that question.  

RE question on moving to introduction section, I think, the intro is long enough and I would like to avoid it. @billwuqin has a comment of creating an Operational Consideration Section #38. Maybe we find a home for this text there? Create a sub-section (name TBD)?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -129,10 +129,16 @@ network to signal the maximum allowable bit rate, and reduce CPU
...
+While endpoints send SCONE packets as frequently as they see fit to ensure
+reliable delivery, the Network Element makes independent decisions on how
+frequently to update those packets to mitigate packet loss. This decision
+relies on operational considerations such as CPU load and the nature of
+the network policies. A network enforcing dynamic policies will prioritize
+timely updates to minimize the delay in activating the advised bit-rate after
+a packet loss. Conversely, a network enforcing fixed, subscription-based
+policies that do not change over SCONE timescales can safely scale back the
+Network Element update frequency to conserve CPU resources, as timely recovery
+from packet loss is less critical.
```
mirjak: Not sure I fully understand the relation to packet loss here. Can you further explain?

smishra1200: @mirjak, Reading back, yes the text is not only confusing but there is no real need to discuss packet loss. Here is the *clean* text to replace part of the section. Also, I will also remove verbiage from the first paragraph that has language from the protocol document. I will add a PR to correct it.

Here is the part of the paragraph
"...A network enforcing dynamic policies might prioritize updating SCONE packets immediately upon a policy trigger to minimize the application's reaction time to the new limit. Conversely, a network enforcing fixed, subscription-based policies can safely scale back its update frequency to conserve CPU resources, provided it still updates SCONE packets periodically (e.g., every 20 to 30 seconds). This periodic update frequency ensures that the throughput advice reliably reaches the endpoint and does not inadvertently expire across the standard monitoring period due to normal packet loss."
."

billwuqin (commented):
Why both Section on Retransmission Rate of Advised Bit-Rate and Section on Standards Compliance get updated in one PR?

ihlar (approved):
Thanks for the update, it looks good to me.

PR #37, opened 2026-04-23
SCONE and (MASQUE) tunnels

mirjak:


draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +190,35 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
 would benefit from harmonizing multiple congestion signaling methods by policy or scope
 deployments to avoid conflicting feedback.
 
+# Using SCONE in and for QUIC tunnels using MASQUE
```
```suggestion
# Using SCONE in and for QUIC tunnels in MASQUE
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +190,35 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
...
+such as a MASQUE client, treats this advice as an upper bound on tunnel capacity
+and has to decide how to share that capacity across tunneled connections.
+
+If SCONE is used on the inner traffic as well as for the tunnel QUIC connection, the MASQUE client
+might receive two different throughput advise values. The client should take the minimum of the throughput values.
+If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the client needs to
+act as a network element and ensure the value in the SCONE packet of the outer QUIC connection
+is translated to an inner
+SCONE packet in order to be visible to the transport connection endpoints. Again in this case
+the MASQUE client has to decide how to share the capacity among all tunneled connections.
```
mirjak: Adding more explanation about the simple approach to simply set the outer value to the inner scone packet:
```suggestion
the MASQUE client has to decide how to share the capacity among all tunneled connections.
Note that, as describe in the SCONE protocol specification, the throughput advise
has an intentionally unspecified scone (see {{Section 3.2 of I-D.ietf-scone-protocol}}
and as such provides only an upper limit but does not guarantee that the throughput is achievable
and therefore could be enforced at a different scope. Therefore, directly providing the outer values
in the inner SCONE packet is a valid approach but it might depend strongly on the inner traffic mix
if the benefits of SCONE can be realized.
```

smishra1200: @mirjak, this is a great addition! The text nicely captures discussion on it from IETF 125 that MASQUE proxies can simply act as SCONE network elements without needing new signaling.

I have one minor comment on the terminology, we have normalized terminology in the PRs to say either per flow of packets on the same UDP address tuple, or in aggregate across multiple QUIC flows, so perhaps we can align here as well?

Oh, need to add the citation for {{?I-D.schinazi-masque-proxy}} in the informative reference block. 
I-D.schinazi-masque-proxy:
    target: https://datatracker.ietf.org/doc/draft-ietf-masque-connect-udp/
    title: "Proxying UDP in HTTP"
    author:
      - name: D. Schinazi
    date: 2022-08
    seriesinfo: "Internet-Draft, draft-ietf-masque-connect-udp, Work in Progress"

billwuqin: Based on what I heard in the interim, need to re-focus on key challenge to using the SCONE protocol within and for QUIC tunnels in MASQUE.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +190,35 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
...
 deployments to avoid conflicting feedback.
 
+# Using SCONE in and for QUIC tunnels in MASQUE
+
+When traffic is tunneled on the network path, such as when using MASQUE {{?I-D.schinazi-masque-proxy}},
+SCONE maybe be carried on the inner traffic but can also be used on a QUIC-based tunnel itself.
+
+If the inner traffic is SCONE-capable, the tunnel endpoint itself
+can be a SCONE-capable network element as it has visibility into the demultiplexed tunneled traffic.
+In this case, throughput advice can used be to manage individual tunneled flows.
```
martinthomson: General comment here rather than a specific one: don't use the word "manage" for what a network element does with SCONE.  It's a request (albeit with teeth) to endpoints.  Saying "manage" implies a degree of control that doesn't exist.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +190,35 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
...
+
+If the MASQUE tunnel itself uses a SCONE-enabled QUIC connection,
+the SCONE-capable network element is on the path between the MASQUE client and the MASQUE proxy.
+In this case, the tunnel is the only flow visible to the network element
+and as such SCONE throughput advice applies to the tunnel as an aggregate QUIC connection.
+The advice reflects the maximum achievable throughput for traffic carried over the tunnel,
+independent of how many tunneled flows are multiplexed within it. The receiving tunnel endpoint,
+such as a MASQUE client, treats this advice as an upper bound on tunnel capacity
```
martinthomson: Not as bad as "manage", but "upper bound on capacity" implies more certainty than exists in practice.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +190,35 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
...
+and has to decide how to share that capacity across tunneled connections.
+
+If SCONE is used on the inner traffic as well as for the tunnel QUIC connection, the MASQUE client
+might receive two different throughput advise values. The client should take the minimum of the throughput values.
```
martinthomson:
```suggestion
might receive two different throughput advice values. The client should take the minimum of the throughput values.
```

From the perspective of the endpoint, the advice doesn't apply to the same thing at all, so this is not a good framing.

You should instead say that one applies to the tunnel (as above) and the other applies to the tunneled flow.  It should be obvious, of course, that the capacity of the tunnel will constrain how much the tunneled flow can use, so it makes sense to reduce the value associated with the tunneled flow to consider any lower rate received for advice on the tunnel.  Your framing could be read as applying in the opposite direction.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +190,35 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
...
+If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the client needs to
```
martinthomson:
```suggestion
If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the client can
```

"needs to" is a bit strong.  Obviously, it is better if the advice that is received on the tunnel is used to update advice for the tunneled flow, but this reads like a "MUST, but we can't say MUST here".  Instead, I'd prefer that this make the case for transferring advice: 

> If an endpoint forwards the tunneled flow to another entity, it can improve the advice that the ultimate endpoint of that flow receives by updating the throughput advice to take the advice it received for the tunnel into account.

Then an aside, perhaps:
> Note that this could end up extending the time that reduced throughput advice is applied by an endpoint, due to the time that elapses between the receipt of advice on the tunnel and the next opportunity to update advice on the tunneled flow.  This can have the effect of delaying the response of endpoints by as much as double the monitoring period.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +190,35 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
...
+act as a network element and ensure the value in the SCONE packet of the outer QUIC connection
+is translated to an inner
+SCONE packet in order to be visible to the transport connection endpoints. Again in this case
+the MASQUE client has to decide how to share the capacity among all tunneled connections.
+
+Using SCONE on a MASQUE QUIC tunnel connection can enable throughput advise also for tunnelled
+non-QUIC traffic, however, this can only be applied if there is a interface or mechanism in the client
+that provides the throughput advise to the application.
```
martinthomson:
```suggestion
Using SCONE on a MASQUE QUIC tunnel connection can enable throughput advice also for tunneled
non-QUIC traffic, however, this can only be applied if there is a interface or mechanism in the client
that provides the throughput advice to the application.
```
(Just spelling)

draft-ietf-scone-applicability-manageability.md
mirjak:
```suggestion
In this case, throughput advice can be applied to individual tunneled flows.
```

draft-ietf-scone-applicability-manageability.md
mirjak:
```suggestion
such as a MASQUE client, can apply this advice as a potential upper bound on tunnel capacity
```

draft-ietf-scone-applicability-manageability.md
martinthomson: "can be applied"?  or "can apply"

Your text, BTW, feel free to merge suggestions.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,65 @@ isolation. Metrics of interest include:
...
+If SCONE is used on the inner traffic as well as for the tunnel QUIC connection, the MASQUE client
+might receive two different throughput advise values. The client should take the minimum of the throughput values.
+If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the client needs to
+act as a network element and ensure the value in the SCONE packet of the outer QUIC connection
+is translated to an inner
+SCONE packet in order to be visible to the transport connection endpoints. Again in this case
+the MASQUE client has to decide how to share the capacity among all tunneled connections.
+
+Using SCONE on a MASQUE QUIC tunnel connection can enable throughput advise also for tunnelled
+non-QUIC traffic, however, this can only be applied if there is a interface or mechanism in the client
```
anooptomar29: "And this may be possible if there is only one tunneled non-quic traffic".. Should we add this also to clarify this further ?

mirjak: I think these same advice as above applies here where you would need to decide how to slice and dice the advised rate among multiple flows.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,65 @@ isolation. Metrics of interest include:
...
+independent of how many tunneled flows are multiplexed within it. The receiving tunnel endpoint,
+such as a MASQUE client, can apply this advice as a potential upper bound on tunnel capacity
+and has to decide how to share that capacity across tunneled connections.
+
+If SCONE is used on the inner traffic as well as for the tunnel QUIC connection, the MASQUE client
+might receive two different throughput advise values. The client should take the minimum of the throughput values.
+If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the client needs to
+act as a network element and ensure the value in the SCONE packet of the outer QUIC connection
+is translated to an inner
+SCONE packet in order to be visible to the transport connection endpoints. Again in this case
```
```suggestion
If an endpoint forwards the tunneled flow to another entity, it can update the advice that the ultimate endpoint of that flow receives by taking the throughput advice it received for the tunnel into account.

Note that this could end up extending the time that updates throughput advice is applied by an endpoint, due to the time that elapses between the receipt of advice on the tunnel and the next opportunity to update advice on the tunneled flow. This can have the effect of delaying the response of endpoints by as much as double the monitoring period.

Again in this case
```

draft-ietf-scone-applicability-manageability.md
mirjak: I applied your suggestion below with some wording change to simplify a bit. I know it changes you sentence slightly but I hope it's good enough? Please have a look!

draft-ietf-scone-applicability-manageability.md
mirjak: Still not sure if we want to commit this or not?

draft-ietf-scone-applicability-manageability.md
mirjak: Is this about this sentence or rather the below? Can you propose something?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,65 @@ isolation. Metrics of interest include:
...
+works at application level. SCONE does not replace the need for endpoints to perform
+congestion control or network to provide explicit or implicit congestion signals; rather,
+it complements these mechanisms by providing a variable range for application-level rate
+adaptation. In environments where both are present, SCONE and congestion control mechanisms
+co-exist: congestion control manages the immediate dynamics of the bottleneck link, while
+SCONE informs the application of the maximum sustained rate allowed by policy. Network Operators
+would benefit from harmonizing multiple congestion signaling methods by policy or scope
+deployments to avoid conflicting feedback.
+
+# Using SCONE in and for QUIC tunnels in MASQUE
```
mirjak: Is this any better?
```suggestion
# Using SCONE in tunnels and for QUIC-based tunnels in MASQUE
```

draft-ietf-scone-applicability-manageability.md
billwuqin: If we can have more people review and approve PR, I think we can merge. thanks!

draft-ietf-scone-applicability-manageability.md
anooptomar29: Sounds good.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,65 @@ isolation. Metrics of interest include:
...
+the SCONE-capable network element is on the path between the MASQUE client and the MASQUE proxy.
+In this case, the tunnel is the only flow visible to the network element
+and as such SCONE throughput advice applies to the tunnel as an aggregate QUIC connection.
+The advice reflects the maximum achievable throughput for traffic carried over the tunnel,
+independent of how many tunneled flows are multiplexed within it. The receiving tunnel endpoint,
+such as a MASQUE client, can apply this advice as a potential upper bound on tunnel capacity
+and has to decide how to share that capacity across tunneled connections.
+
+If SCONE is used on the inner traffic as well as for the tunnel QUIC connection, the MASQUE client
+might receive two different throughput advise values. The client should take the minimum of the throughput values.
```
martinthomson:
```suggestion
might receive two different throughput advice values.
Each piece of advice applies independently to the flow on which it is received.
```

martinthomson (commented):
Overall, this seems like too much.  The one sentence I made a suggestion on below is probably all that is needed.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,60 @@ isolation. Metrics of interest include:
...
+works at application level. SCONE does not replace the need for endpoints to perform
+congestion control or network to provide explicit or implicit congestion signals; rather,
+it complements these mechanisms by providing a variable range for application-level rate
+adaptation. In environments where both are present, SCONE and congestion control mechanisms
+co-exist: congestion control manages the immediate dynamics of the bottleneck link, while
+SCONE informs the application of the maximum sustained rate allowed by policy. Network Operators
+would benefit from harmonizing multiple congestion signaling methods by policy or scope
+deployments to avoid conflicting feedback.
+
+# Using SCONE in tunnel and for QUIC-based tunnels in MASQUE
```
```suggestion
# Using SCONE in tunnels and for QUIC-based tunnels in MASQUE
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,60 @@ isolation. Metrics of interest include:
...
+the SCONE-capable network element is on the path between the MASQUE client and the MASQUE proxy.
+In this case, the tunnel is the only flow visible to the network element.
+
+Where flows are tunneled, throughput advice applies to the flow to which it was attached only.
+This might mean that a QUIC flow that is tunneled in QUIC
+might have multiple, independent pieces of throughput advice.
+
+If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the tunnel egress 
+might consider the throughput advice of the outer tunnel when providing throughput advice on
+the inner SCONE-enables flows.
```
gorryfair:
```suggestion
the inner SCONE-enabled flows.
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,60 @@ isolation. Metrics of interest include:
...
+adaptation. In environments where both are present, SCONE and congestion control mechanisms
+co-exist: congestion control manages the immediate dynamics of the bottleneck link, while
+SCONE informs the application of the maximum sustained rate allowed by policy. Network Operators
+would benefit from harmonizing multiple congestion signaling methods by policy or scope
+deployments to avoid conflicting feedback.
+
+# Using SCONE in tunnels and for QUIC-based tunnels in MASQUE
+
+When traffic is tunneled on the network path, such as when using MASQUE {{?I-D.schinazi-masque-proxy}},
+SCONE maybe be carried on the inner traffic but can also be used on a QUIC-based tunnel itself.
```
gorryfair:
```suggestion
SCONE can be carried on the inner traffic but can also be used on a QUIC-based tunnel itself.
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,60 @@ isolation. Metrics of interest include:
...
+
+If the inner traffic is SCONE-capable, each tunnel endpoints
```
gorryfair:
```suggestion
If the inner traffic is SCONE-capable, each tunnel endpoint
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,60 @@ isolation. Metrics of interest include:
...
+can be a SCONE-capable network element. In this case, as the tunnel endpoints have visibility into
+the demultiplexed tunneled traffic, throughput advice can be applied to individual tunneled flows.
+
+If a MASQUE tunnel uses a SCONE-enabled QUIC connection,
+the SCONE-capable network element is on the path between the MASQUE client and the MASQUE proxy.
+In this case, the tunnel is the only flow visible to the network element.
```
gorryfair, line 204: only flow visible isn't that clear -- I think you mean the inner flow(s) are not visible?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,60 @@ isolation. Metrics of interest include:
...
+Where flows are tunneled, throughput advice applies to the flow to which it was attached only.
+This might mean that a QUIC flow that is tunneled in QUIC
+might have multiple, independent pieces of throughput advice.
+
+If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the tunnel egress 
+might consider the throughput advice of the outer tunnel when providing throughput advice on
+the inner SCONE-enables flows.
+
+Using SCONE on a MASQUE QUIC tunnel connection can enable throughput advice also for tunneled
+non-QUIC traffic, however, this can only be applied if there is a interface or mechanism in the
```
gorryfair:
```suggestion
non-QUIC traffic, however, this can only be applied if there is an interface or mechanism in the
```

gorryfair (commented):
These comments are to tidy the text - not upon the usefulness of the text.

draft-ietf-scone-applicability-manageability.md
mirjak: yes. We could say "only the QUIC connection of the tunnel is visible to the network element which might carry multiple end-to-end flow." better?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -248,93 +163,60 @@ isolation. Metrics of interest include:
...
+can be a SCONE-capable network element. In this case, as the tunnel endpoints have visibility into
+the demultiplexed tunneled traffic, throughput advice can be applied to individual tunneled flows.
+
+If a MASQUE tunnel uses a SCONE-enabled QUIC connection,
+the SCONE-capable network element is on the path between the MASQUE client and the MASQUE proxy.
+In this case, the tunnel is the only flow visible to the network element.
+
+Where flows are tunneled, throughput advice applies to the flow to which it was attached only.
+This might mean that a QUIC flow that is tunneled in QUIC
+might have multiple, independent pieces of throughput advice.
```
```suggestion
In the tunnel endpoint is also the application endpoint,
this might mean that a QUIC flow that is tunneled in QUIC
might have multiple, independent pieces of throughput advice.
```

draft-ietf-scone-applicability-manageability.md
gorryfair: That sounds clearer to me (and was the property that I had expected).

mirjak:
> Trying to parse what is actually said, I'm not sure what guidance is actually provided by the current text that is beyond saying that you can see any outer header, and a tunnel endpoint has to figure out what to do when stripping the tunnel encaps.

Yes, there is not much guidance left but maybe there is much we guidance we can give here. One point, as you say, it explicitly point out that the outer and inner signals net to be considered as independent. This other small piece of advice is this: "the tunnel egress 
might consider the throughput advice of the outer tunnel when providing throughput advice on
the inner SCONE-enabled flows." But yes, that is very vague. I'd be all in to provide something more concrete but I got the feeling that this is where the group converged on...?

draft-ietf-scone-applicability-manageability.md
mirjak:
```suggestion
In this case, only the QUIC connection of the tunnel is visible to the network element which might carry multiple end-to-end flows.
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -341,6 +341,32 @@ SCONE seamlessly integrates into existing architectures without requiring new tu
...
+
+If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the tunnel egress
```
```suggestion
If the MASQUE client is not also the endpoint of the end-to-end QUIC connection, the tunnel egress
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -341,6 +341,32 @@ SCONE seamlessly integrates into existing architectures without requiring new tu
 or data paths to be established.
 
 
+## Using SCONE in tunnels and for QUIC-based tunnels in MASQUE
```
anooptomar29: For extreme clarity could we add a note to clearly state that this is not about SCONE versus Masque, but rather using SCONE for masque tunneled traffic.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -341,6 +341,32 @@ SCONE seamlessly integrates into existing architectures without requiring new tu
...
+
+When traffic is tunneled on the network path, such as when using MASQUE {{?I-D.schinazi-masque-proxy}},
+SCONE can be carried on the inner traffic but can also be used on a QUIC-based tunnel itself.
```
zaheduzzaman: SCONE is always tunneled for Masque.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -341,6 +341,32 @@ SCONE seamlessly integrates into existing architectures without requiring new tu
...
+
+If the inner traffic is SCONE-capable, each tunnel endpoint
+can be a SCONE-capable network element. In this case, as the tunnel endpoints have visibility into
+the demultiplexed tunneled traffic, throughput advice can be applied to individual tunneled flows.
```
zaheduzzaman: Here we should say, the advice is applied only to the flows the scone packet is received.

zaheduzzaman (commented):
Looks OK to me, however, there is QUIC aware proxying and we have not mention of that here.

Issue #38, opened 2026-04-24
Add Operational Consideration Section to align with RFC5706bis


billwuqin:

smishra1200:
Great point. An option is to like add an "Operational Considerations" section that aligns with RFC5706bis guidelines. 

This section can piece together the operational lifecycle for network providers, such as, 
1) how throughput constraints are determined (policy/configuration);
2) how signaling failures are detected (fault management)
3) and how compliance is monitored (performance management)

Just some preliminary thoughts. Would his align with what you are suggesting? 

We will just need to do rejig in the draft to get this aligned to your point.

smishra1200:
@billwuqin, I have submitted PR #39 as a start. I have added a sub-section, Determining Throughput Constraints, that covers topics such as, Subscriber Policies and Data Plans, Application and Video Optimization Policies, Dynamic Network Conditions and Capacity and Load Management. That said, we will now have some duplicative text that is elsewhere in the document. That can be easily handled once the WG converges on what is the right text and where it belongs. I have also started another sub-section as a place holder "Detecting Signaling Failures" but need to work on the text (do I hear a volunteer :) ). and also potentially other operational considerations.

billwuqin:
Good start, please refer to https://github.com/IETF-OPS-DIR/Review-Template/blob/main/opsdir-review-template.md
for more guidance on what are needed.
This template is designed for reviewers, but I think it will be also useful for authors of I-Ds.

Issue #21, opened 2026-04-14
Explain how throughput constraints are determined


billwuqin:
e.g., do you measure actual bit rate per flow, do you observes its local radio and network conditions, do you subscription-based policy?

zaheduzzaman:
The constrains can be very network operations dependent, i.e. the network operators decided and the SCONE protocol does not care about how the rate is calculated, it could be some policy based  or some other more dynamic approched. The document now says this is out of scope.

smishra1200:
To add to what @zaheduzzaman wrote, please see: [https://github.com/ietf-wg-scone/appman/pull/34/changes/9bf0ff8c9b3f87b36fc15264b72be14b17102bff#:~:text=longer%20time%20period.-,The,are%20out%20of%20scope%20for%20this%20document.,-199](url)

billwuqin:
> The constrains can be very network operations dependent, i.e. the network operators decided and the SCONE protocol does not care about how the rate is calculated, it could be some policy based or some other more dynamic approched. The document now says this is out of scope.

I think SCONE protocol focus on how throughput advice can be communicated while Applicability and Management might choose to    document missing pieces, e.g., how throughput constraints are determined, some examples will help, otherwise, I feel many network operators don't how the complete solution can be implemented. Many missing pieces leave to implementation specific sometimes doesn't help, make sense?

smishra1200:
Fair point. While the specific mathematical calculation of the rate is implementation-specific and depends on the operator's business logic, there is merit to have some text in Appman to provide examples of the inputs and policies that determine these constraints.  Maybe add a new section "Determining Throughput Constraints" to explicitly outline examples of how throughput advice is determined, such as subscriber data thresholds (communicated via PCF/SMF), video optimization policies, and dynamic capacity changes like RAT handovers or equipment faults etc?

gorryfair:
I think this can only be examples - there must be millions of ways to do this, and it needs to say this is simply some diverse ways to make this work, but I'm interested to see some examples of how this might work.

smishra1200:
@gorryfair wrote:

> I think this can only be examples - there must be millions of ways to do this, and it needs to say this is simply >some diverse ways to make this work, but I'm interested to see some examples of how this might work.

[...]

PR #39, opened 2026-04-28
Add Operational Considerations Section

smishra1200:
Issue #38: Add a new Operational Considerations Section

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -170,10 +170,6 @@ adjust policies. Due to flow awareness, such mechanisms are typically
 implemented in a SCONE Network Element but may also be implemented
 elsewhere in the network.
 
-## Standards Compliance
-SCONE signaling is expected to traverse the existing data path associated
-with the UDP 4-tuple flow for which the Network Element intends to send the advisory bit-rate.
-
 ## Interworking with Other Congestion Management Mechanisms
 SCONE is distinct from transport-level congestion control mechanisms, such as
 Explicit Congestion Notification (ECN) or Low Latency, Low Loss, and Scalable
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -190,6 +186,65 @@ SCONE informs the application of the maximum sustained rate allowed by policy. N
 would benefit from harmonizing multiple congestion signaling methods by policy or scope
 deployments to avoid conflicting feedback.
 
+# Operational Considerations
```
mirjak, line 189: Can we maybe just have the subsection and not this intro text as the whole document is basically on operational considerations?

britram: Point, though my understanding is that OPSDIR will look for this section.

britram: (We might want some language here pointing out that the whole doc is basically operational considerations though)

smishra1200: I'm fine either way so please decide which way we want to go and then I will use my scalpel to work through the section.

gorryfair: As a contributor to SCONE (not as AD), I liked the idea of a short operational considerations section that says something like: SCONE requires configuration in servers and clients; network devices that forward QUIC packets ought to also forward packets including SCONE. Specific network devices can implement and configure the SCONE Network Element function. These aspects are discussed in /this document/. ... And I heard a suggestion that the WG added a similar statement to the protocol spec that said about the same but pointed to the /applicability and manageability document/.  It's one approach that might be clear?

smishra1200: @gorryfair, here is what we currently have:

**4. Operational Considerations**
Although the core SCONE protocol specifies how throughput advice is communicated to endpoints, deploying SCONE requires network operators to integrate this signaling into their existing operational, policy, and management frameworks. To implement a complete and effective solution, operators must address several operational lifecycle phases, including **policy configuration, fault detection, and performance monitoring**.

This section outlines the practical operational considerations and manageability requirements for deploying SCONE within an operator network, ensuring that SCONE Network Elements can reliably determine, deliver, and monitor throughput advice while addressing real-world network challenges.

And then the subsections are proposed (but not fully developed except for 4.1. Determining Throughput Constraints).

I see there are questions, one if section on the need for a section (sec 4) and if agreeable, what suitable subsections to address. At this points, we are awaiting further decision on the present introductory text in 4 and the 4.1 subsection (which starts to address "policy configuration" part in 4.1). Some of the feedback and discussions at the last interim was that going into operator policy may be challenging as this will vary by operator and presumably this is not exhaustive. 

I would like to get some clarity on this whole section so we can take action accordingly. 

For completeness here, I am adding the text as included in the PR above for section 4.1 

**4.1. Determining Throughput Constraints**
The specific algorithms used to calculate throughput advice are highly dependent on an operator's internal business logic and network architecture. In practice, however, these constraints are generally determined by mapping established network policies and real-time conditions into SCONE signals. To implement a complete solution, operators typically derive their throughput advice from one or more of the following operational inputs: 

**Subscriber Policies and Data Plans:** Rate limits may be triggered when a subscriber reaches specific data plan thresholds or usage caps. For example, in a 5G architecture, the User Plane Function (UPF) acting as the SCONE network element receives Quality of Service (QoS) and charging policy rules from the Policy Control Function (PCF) via the Session Management Function (SMF). Similarly, in a 4G network, the Packet Data Network Gateway (P-GW) applies policies received from the Policy and Charging Rules Function (PCRF). Beyond mobile networks, this applies equally to wireline broadband networks enforcing policies at a Broadband Network Gateway (BNG), or enterprise Wi-Fi networks where rate limits are enforced at an Access Point (AP) or wireless LAN controller.

**Application-Specific Policies:** Operators may configure specific maximum bit-rates allowed for certain types of traffic based on the user's subscription tier or device type. This includes video optimization policies for adaptive bitrate video, as well as traffic shaping policies applied to low-priority bulk downloads (such as background software updates).

**Dynamic Network Conditions:** Constraints can be updated dynamically due to changes in the access network type, such as Radio Access Technology (RAT) handovers when a user moves between 5G and 4G coverage areas.
Capacity and Load Management: During periods of unusually high usage, persistent overuse, or temporary equipment faults, the network element might signal a temporarily reduced throughput advice to manage shared capacity and guide application network usage.

By utilizing these policy rules and network conditions, operators establish the inputs necessary for the SCONE network element to dynamically determine and apply the appropriate throughput advice for a given flow.

billwuqin:
@smishra1200  I think Brian and Gorry have already provided suggestions in the earlier discussion for you.  My feeling #21 is about how to configure SCONE network element, it can be separated out from operational consideration, since operational consideration is not limited to configuration consideration, it also include mointoring, measurment, etc. In addition, I feel PR#39 is not generic enough, Gorry also provide similar suggestion.
Here is the way forward, you may consider to adding a new indepedent section to discuss how throughput constraints are determined.
And then you have a short operational document, to clarify the whole document is about operational consideration or management consideration
"
xxx Operational Considerations

SCONE requires configuration in servers and clients; network devices that forward QUIC packets ought to also forward packets including SCONE. Specific network devices can implement and configure the SCONE Network Element function. These aspects are discussed in this document.
"
With this short short Operational Considerations section, we can request OPS-DIR review to identify what is missing, make sense?

PR #42, opened 2026-06-11
Add Determining Throughput Constraints section per PR #39 (Issue #21)

smishra1200:
Addresses #21. Adds a new Determining Throughput Constraints subsection after Flow Awareness and Per-Flow Signaling, with illustrative, technology-neutral inputs an operator may use to determine throughput advice. Split out from PR #39 per the restructure discussed there.


billwuqin (approved):
Thanks for addressing my comment, the proposed change is much better than previous one.

PR #55, opened 2026-07-12
Issue #48, Add diagnosis guidance for throughput advice not followed

smishra1200:
Issue #48,   Add a paragraph at the end of the Monitoring and Logging section to help operators distinguish between a delivery failure and application non-compliance when throughput advice does not appear to be followed. References Section 3.5 and Section 7.2 of I-D.ietf-scone-protocol for the advisory signal behavior and monitoring guidance respectively.

Added Text:
When throughput advice does not appear to be followed, it is useful to determine whether the advice reached the application at all, or whether the application received it but did not act on it. Operators can narrow the diagnosis by correlating the rate of SCONE advisories issued at the network element against the observed per-flow throughput over the following two monitoring periods. If the network element successfully updates traversing SCONE packets during that window but the flow's throughput does not change, it indicates either that the packets were dropped downstream before reaching the application, or that the application received the advice but did not act upon it, as SCONE is an advisory signal per Section 3.5 of I-D.ietf-scone-protocol. Conversely, if the network element stops observing traversing SCONE packets arriving from the sender, this suggests either an upstream delivery failure or the endpoint is no longer sending SCONE packets on that flow. Recording both the timestamps of issued advisories and the subsequent per-flow throughput measurements in the logging infrastructure makes this correlation possible, aligning with the monitoring guidance in Section 7.2 of
I-D.ietf-scone-protocol.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -234,6 +234,28 @@ isolation. Metrics of interest include:
 - Correlation between SCONE advisories and user-plane throughput changes
 - Error conditions where SCONE signaling fails to reach the intended endpoints
 
+When throughput advice does not appear to be followed, it is
+useful to determine whether the advice reached the application
+at all, or whether the application received it but did not act
+on it. Operators can narrow the diagnosis by
+correlating the rate of SCONE advisories issued at the network
+element against the observed per-flow throughput over the
+following two monitoring periods. If the network element
+successfully updates traversing SCONE packets during that window
+but the flow's throughput does not change, it indicates either
+that the packets were dropped downstream before reaching the
+application, or that the application received the advice but did
+not act upon it, as SCONE is an advisory signal per
+{{Section 3.5 of I-D.ietf-scone-protocol}}. Conversely, if the
+network element stops observing traversing SCONE packets
+arriving from the sender, this suggests either an upstream delivery
+failure or the endpoint is no longer sending SCONE packets on that flow.
+Recording both the timestamps of issued advisories and the subsequent
+per-flow throughput measurements in the logging infrastructure
+makes this correlation possible, aligning with the monitoring
+guidance in {{Section 7.2 of I-D.ietf-scone-protocol}}.
+
+
 ## Conformance Monitoring
 Networks that choose to provide SCONE throughput advice can implement mechanisms to
 monitor QUIC flows and measure conformance to the advised bit-rate, either per flow of
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -250,7 +250,7 @@ not act upon it, as SCONE is an advisory signal per
 network element stops observing traversing SCONE packets
 arriving from the sender, this suggests either an upstream delivery
 failure or the endpoint is no longer sending SCONE packets on that flow.
-Recording both the timestamps of issued advisories and the subsequent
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -234,6 +234,28 @@ isolation. Metrics of interest include:
...
+When throughput advice does not appear to be followed, it is
+useful to determine whether the advice reached the application
+at all, or whether the application received it but did not act
+on it. Operators can narrow the diagnosis by
+correlating the rate of SCONE advisories issued at the network
+element against the observed per-flow throughput over the
+following two monitoring periods. If the network element
+successfully updates traversing SCONE packets during that window
+but the flow's throughput does not change, it indicates either
+that the packets were dropped downstream before reaching the
```
martinthomson:
```suggestion
that all updated packets were dropped downstream before reaching the
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -234,6 +234,28 @@ isolation. Metrics of interest include:
...
+application, or that the application received the advice but did
+not act upon it, as SCONE is an advisory signal per
+{{Section 3.5 of I-D.ietf-scone-protocol}}. Conversely, if the
+network element stops observing traversing SCONE packets
+arriving from the sender, this suggests either an upstream delivery
+failure or the endpoint is no longer sending SCONE packets on that flow.
+Recording both the timestamps of issued advisories and the subsequent
```
martinthomson:
```suggestion
Recording both the timestamps of updates to packets and the subsequent
```

martinthomson (approved):

ihlar (approved):

draft-ietf-scone-applicability-manageability.md
```diff
@@ -234,6 +234,28 @@ isolation. Metrics of interest include:
...
 
+When throughput advice does not appear to be followed, it is
+useful to determine whether the advice reached the application
+at all, or whether the application received it but did not act
+on it. Operators can narrow the diagnosis by
+correlating the rate of SCONE advisories issued at the network
+element against the observed per-flow throughput over the
+following two monitoring periods. If the network element
+successfully updates traversing SCONE packets during that window
+but the flow's throughput does not change, it indicates either
```
gorryfair, line 245: ... I am worried about /throughput does not change/: I suggest 'throughput' is the wrong word - this should measure if the rate averaged over 67 secs exceeds the advised value.

smishra1200, line 245:
```suggestion
but the flow's average bitrate does not change during the monitoring period, it indicates either
```

smishra1200: @gorryfair does the suggestion address your comment_

anooptomar29: @smishra1200 To be consistent with the  terminology used in the "Conformance Monitoring" section ( line number 283) I would replace "monitoring period" by "measurement and averaging period" .

smishra1200: @anooptomar29, addressed in #61. 
replaces "monitoring periods" with "measurement and averaging periods" per your comment above. New PR since #55 is already merged. Please review and approve.

anooptomar29: @smishra1200 I reviewed the changes in #61 and I realized that at this place usage of "monitoring period" is appropriate. So you may pls ignore my comment.

PR #62, opened 2026-07-28
Merge sections on reliability, frequency of updates and dynamic updates

mirjak:
This merge aims to avoid redundancy and thereby make operator advice more clear without cross references.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -209,7 +209,7 @@ As SCONE signaling is decided by the
 application endpoint, it is critical to understand for network operator that the frequency of traversing SCONE
 packets varies depending on the specific application type and its traffic
 characteristics rather than expecting a predictable or uniform signaling cadence.
-However, a SCONE-enabled network element does not need to update every SCONE packet but 
+However, a SCONE-enabled network element does not need to update every SCONE packet but
 can implement its own update interval and then wait for the next SCONE packet.
 It is recommended to updates advice in SCONE packets at least twice per the
 67-second monitoring period ({{Section 5.2 of I-D.ietf-scone-protocol}},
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -196,43 +196,44 @@ policy decisions. Network element implementations can provide the
...
-
-## Dynamic Updates
-Target throughput advice can change dynamically while a flow is active, for example when a
-subscriber reaches a data threshold or a network policy changes. When this happens, the network
-element can update the next traversing SCONE packet with the new throughput advice (see
-{{Section 9.2 of I-D.ietf-scone-protocol}}).
+With more dynamic network policies,
+the target throughput advice can change while a flow is active, for example when a
+subscriber reaches a data threshold. When this happens, the network
+element might prioritize updating updating the next traversing SCONE packet (see
```
mirjak, line 235:
```suggestion
element might prioritize updating the next traversing SCONE packet (see
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -196,43 +196,44 @@ policy decisions. Network element implementations can provide the
...
+Further, loss of a SCONE packet impacts the update frequency.
+The reliable delivery of throughput advice relies on the periodic
+sending of SCONE packets by application endpoints.
 
-A SCONE-enabled network element updates advice in SCONE packets at least twice per the
+As SCONE signaling is decided by the
+application endpoint, it is critical to understand for network operator that the frequency of traversing SCONE
+packets varies depending on the specific application type and its traffic
+characteristics rather than expecting a predictable or uniform signaling cadence.
+However, a SCONE-enabled network element does not need to update every SCONE packet but
```
anooptomar29, line 212: In case n/w element does not update every pkt sent  by server endpoint ( with rate signal set to 127) then client app-end point would receive pkt with 127 and it may consider it as no-limit. So best it that n/w element updates every pkt. To reduce the load on n/w , server endpoint  can reduced the freq of scone pkt sending.

mirjak: Not that's not what we specified in the spec. 127 doesn't overwrite a valid signal; but the signal is expired after one monitoring interval.

PR #70, opened 2026-07-29
Merge Reliability and Frequency of Updates, keep Dynamic Updates separate (alternative to #62)

smishra1200:
Alternative to #62, submitted for comparison, not intended to merge
until we have compared it against #62 and settled on a direction. See
the heads-up comment on #62 for context.

Merges Reliability and Mitigating Packet Loss and Frequency of Updates
into one section, retitled "Reliability of Throughput Advice Delivery"
(drops "Mitigating Packet Loss" from the title, agreeing with @mirjak ). 

Frequency of Updates carried no independent operator guidance on its own, 
so its one fact (cadence is endpoint-controlled) becomes the opening context 
for why the network element needs its own update interval, rather than a
separate section restating it.

Keeps Dynamic Updates as its own section, since the trigger-response
guidance there (act promptly on a threshold or policy change) is
distinct in kind from the steady-state cadence guidance above it.
Restores a second trigger example, a network-wide policy taking effect
while a flow is already active, alongside the data-threshold example.

No change to any other section.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -215,8 +215,8 @@ advice goes unheard during any single window. A network enforcing fixed,
 subscription-based policies can typically rely on this baseline, since the
 applicable policy for a flow does not usually change once established. In
 practice, the network element can only update advice when a SCONE packet
-is available to modify. If none arrives once the chosen interval has
-elapsed, the network element waits for the next one, and that arrival is
+is available to modify. When the interval has
+elapsed, the network element waits for the next SCONE packet, and that arrival is
 what sets the real interval between updates.
 
 ## Sending Updates for Dynamic Target Throughput Changes
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -221,7 +221,7 @@ what sets the real interval between updates.
 
 ## Sending Updates for Dynamic Target Throughput Changes
 Target throughput advice can change while a flow is active, for example when a
-subscriber reaches a data threshold, or when a network policy or capacity
+subscriber reaches a data threshold or when the radio access technology or capacity
 allocation changes while the flow is already established. When this happens,
 the network element prioritizes updating the next traversing SCONE packet
 promptly, bypassing its scheduled periodic update interval, to minimize the
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -215,22 +215,14 @@ advice goes unheard during any single window. A network enforcing fixed,
 subscription-based policies can typically rely on this baseline, since the
 applicable policy for a flow does not usually change once established.
 
-Whatever interval an operator selects, the network element can only act on
-it when a SCONE packet is actually available to update. If none arrives
-exactly when the operator's interval elapses, the network element has to
-wait for the next one, so the effective update interval in practice is
-set jointly by the operator's chosen setting and the endpoint's own
-packet cadence.
-
-## Sending Updates for Dynamic Target Throughput Changes
+## Dynamic Updates
 Target throughput advice can change while a flow is active, for example when a
 subscriber reaches a data threshold, or when a network-wide policy such as a
 time-of-day or congestion-based limit takes effect while the flow is already
-established. When the target throughput changes, the network element does
-not wait for its own selected update interval to expire. Instead, it updates
-the next traversing SCONE packet as soon as one arrives, to minimize the
-application's reaction time to the new limit (see {{Section 9.2 of I-D.ietf-scone-protocol}}).
-How soon the application sees the change still depends on when the endpoint next sends a SCONE packet,
+established. When this happens, the network element prioritizes updating the
+next traversing SCONE packet promptly, to minimize the application's reaction
+time to the new limit (see {{Section 9.2 of I-D.ietf-scone-protocol}}).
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -196,43 +196,36 @@ policy decisions. Network element implementations can provide the
...
+applicable policy for a flow does not usually change once established. In
+practice, the network element can only update advice when a SCONE packet
+is available to modify. If none arrives once the chosen interval has
+elapsed, the network element waits for the next one, and that arrival is
+what sets the real interval between updates.
+
+## Sending Updates for Dynamic Target Throughput Changes
+Target throughput advice can change while a flow is active, for example when a
+subscriber reaches a data threshold, or when a network policy or capacity
+allocation changes while the flow is already established. When this happens,
```
anooptomar29: pls merge https://github.com/ietf-wg-scone/appman/pull/74/changes  to add or change in radio access technology

draft-ietf-scone-applicability-manageability.md
```diff
@@ -196,43 +196,36 @@ policy decisions. Network element implementations can provide the
...
-subscriber reaches a data threshold or a network policy changes. When this happens, the network
-element can update the next traversing SCONE packet with the new throughput advice (see
-{{Section 9.2 of I-D.ietf-scone-protocol}}).
+## Reliability of Throughput Advice Delivery
+The frequency at which SCONE packets arrive at a network element is set by the
+application endpoint, not the network, and varies by application type and
+traffic pattern ({{Section 7.1 of I-D.ietf-scone-protocol}}, "Applying
+Throughput Advice Signals"). A network element cannot expect a predictable or
+uniform signaling cadence from the traffic itself, and instead decides its own
+update interval within that flexibility.
```
anooptomar29, line 205: Merge https://github.com/ietf-wg-scone/appman/pull/73/changes to add " However in order to reduce the CPU load on network element, application endpoint can send scone packets at lower frequency."

mirjak: Sorry this change doesn't make sense to me. How does an endpoint know what the right frequency is for the network? The endpoint should send at a rater high frequency to ensure that dynamic changes are conveyed reasonable quickly. But endpoints can decide to not update every scone packet in order to reduce load. That's what we decided in the working groups; that part of why we specified the expiry interval.

zaheduzzaman: I agree with Mirja.. on what we agreed in the WG, with the fact that it is not the endpoint decides on the updates on the rate signal , rather the network does  not need to update on each scone packets it see passing through, it would follow the instruction specified in the SCONE spec.

zaheduzzaman (approved):
This PR looks good to me.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -196,43 +196,36 @@ policy decisions. Network element implementations can provide the
...
+subscription-based policies can typically rely on this baseline, since the
+applicable policy for a flow does not usually change once established. In
+practice, the network element can only update advice when a SCONE packet
+is available to modify. If none arrives once the chosen interval has
+elapsed, the network element waits for the next one, and that arrival is
+what sets the real interval between updates.
+
+## Sending Updates for Dynamic Target Throughput Changes
+Target throughput advice can change while a flow is active, for example when a
+subscriber reaches a data threshold, or when a network policy or capacity
```
mirjak: Actually I would prefer to not have network policy here because I think the network policy itself usually doesn't change in the time frame of an active flow. A policy defined the threshold or capacity limits based on certain factors. but the policy itself doesn't easily change.

```suggestion
subscriber reaches a data threshold or when the radio access technology or capacity
```

Also I'm not sure about capacity allocation. What is meant by this?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -196,43 +196,36 @@ policy decisions. Network element implementations can provide the
...
+SCONE packets are not delivered reliably for a variety of reasons. Operators
+may choose to update more frequently than this baseline, at the cost of
+additional network element CPU load, in exchange for a lower chance that
+advice goes unheard during any single window. A network enforcing fixed,
+subscription-based policies can typically rely on this baseline, since the
+applicable policy for a flow does not usually change once established. In
+practice, the network element can only update advice when a SCONE packet
+is available to modify. If none arrives once the chosen interval has
+elapsed, the network element waits for the next one, and that arrival is
+what sets the real interval between updates.
```
mirjak: I guess it will be the normal case that there is not immediately a scone packet when the timer expired, so I suggest to rephrase it more like this:

```suggestion
is available to modify. When the interval has
elapsed, the network element waits for the next SCONE packet, and that arrival is
what sets the real interval between updates.
```

draft-ietf-scone-applicability-manageability.md
anooptomar29: @smishra1200  @mirjak May be I did not phrase it correctly. What I wanted to add is that since it is allowed to the sender to decide on scone pkt frequency it can send it at lower frequency also. Essentially there are some CSPs who wants to n/w to mark every pkt received from server ( dont want to have this logic in n/w element to skip some scone pkts) and to reduce the load on n/w element it is expected that sender ( server) to send pkt at lower frequency (mutually agreed). I want to add this aspect w/o getting into specifics. I will rephrase and submit new PR.

PR #71, opened 2026-07-31
Editorial: clarify data-plan basis in "Subscriber Policies and Data P…

anooptomar29:
Editorial change only — no technical/normative change. In "Determining Throughput Constraints" → "Subscriber Policies and Data Plans", the existing sentence describes throughput advice only in terms of rate limits at a threshold/cap. This rephrasing first states the general case (advice is based on the subscriber's data plan) and then presents rate limits as an additional, more specific case, which reads more clearly.


mirjak (approved):

smishra1200 (approved):
Thanks Anoop, this is much better written sentence.

PR #73, opened 2026-07-31
Frequency of updates section

anooptomar29:
Added text to highlight that if needed end-point can send scone pkts at lower freq ( to reduce CPU load on n/w element if needed)

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -226,7 +226,7 @@ endpoint-generated packets to communicate throughput advice, the frequency of tr
 packets varies depending on the specific application type and its traffic
 characteristics. Consequently, network elements need to be prepared to apply updates to
 traversing packets at highly variable, application-driven intervals rather than expecting a
-predictable or uniform signaling cadence from the network side.
+predictable or uniform signaling cadence from the network side. However in order to reduce the CPU load on network element, application endpoint can send scone packets at lower frequency. 
 
 ## Dynamic Updates
 Target throughput advice can change dynamically while a flow is active, for example when a


draft-ietf-scone-applicability-manageability.md
```diff
@@ -226,7 +226,7 @@ endpoint-generated packets to communicate throughput advice, the frequency of tr
 packets varies depending on the specific application type and its traffic
 characteristics. Consequently, network elements need to be prepared to apply updates to
 traversing packets at highly variable, application-driven intervals rather than expecting a
-predictable or uniform signaling cadence from the network side.
+predictable or uniform signaling cadence from the network side. However in order to reduce the CPU load on network element, application endpoint can send scone packets at lower frequency.
```
mirjak, line 229: As I commented in the other PR, this is not the way it works as the endpoint doesn't know which rate is good for the network element.

smishra1200: @anooptomar29, flagging two things here. 

First, like #74, this targets text that PR #70 rewords entirely ("Frequency of Updates" gets folded into "Reliability of Throughput Advice Delivery"), so this will conflict once #70 merges. 

Second, please see comments from both @mirjak and @zaheduzzaman

anooptomar29: @mirjak @smishra1200 May be I did not phrase it correctly. What I wanted to add is that since it is allowed to the sender to decide on scone pkt frequency it can send it at lower frequency also. Essentially there are some CSPs who wants to n/w to mark every pkt recevied from server ( dont want to have this logic in n/w element to skip some scone pkts) and to reduce the load on n/w element it is expected that sender ( server) to send pkt at lower frequency (mutually agreed). I want to add this aspect w/o getting into specifics. I will rephrase and submit new PR.

PR #75, opened 2026-07-31
Update section multiple scone elements

anooptomar29:
Added text for additional clarity

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -236,12 +236,12 @@ element can update the next traversing SCONE packet with the new throughput advi
 How soon the application sees the change depends on when the endpoint next sends a SCONE packet,
 since the network element cannot originate one.
 
-## Presence of SCONE Network Elements On the Data Path
+## Presence of multiple SCONE Network Elements On the Data Path
 When multiple SCONE-capable network elements are present on the same data path, they operate
 independently, with no synchronization or control-plane coordination required between them.
-Each network element only lowers the rate signal, preserving any lower advice already set by
+Each network element only lowers the rate signal if rate signal in the received scone packet is set to a valid value , preserving any lower advice already set by
 another element on the path, so the endpoint applies the most restrictive advice along the
-path (see {{Section 5.4 of I-D.ietf-scone-protocol}}). This lets operators deploy and manage
+path (see {{Section 5.4 of I-D.ietf-scone-protocol}}).If network element receives scone packet with rate signal set to unknown (127) and if it does not want to restrict the bit-rate then it does not need to update the scone packet. This lets operators deploy and manage
 SCONE network elements independently, without building integration between them.
 
 ## Change of Network Element During an Active Flow


draft-ietf-scone-applicability-manageability.md
```diff
@@ -236,12 +236,12 @@ element can update the next traversing SCONE packet with the new throughput advi
 How soon the application sees the change depends on when the endpoint next sends a SCONE packet,
 since the network element cannot originate one.
 
-## Presence of SCONE Network Elements On the Data Path
+## Presence of multiple SCONE Network Elements On the Data Path
 When multiple SCONE-capable network elements are present on the same data path, they operate
 independently, with no synchronization or control-plane coordination required between them.
-Each network element only lowers the rate signal, preserving any lower advice already set by
+Each network element only lowers the rate signal if rate signal in the received scone packet is set to a valid value , preserving any lower advice already set by
```
mirjak, line 242: This is not a correct change. It changes the meaning of the sentence. Also how can the network element that there is a valid value? I think it just cannot....?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -236,12 +236,12 @@ element can update the next traversing SCONE packet with the new throughput advi
...
 another element on the path, so the endpoint applies the most restrictive advice along the
-path (see {{Section 5.4 of I-D.ietf-scone-protocol}}). This lets operators deploy and manage
+path (see {{Section 5.4 of I-D.ietf-scone-protocol}}).If network element receives scone packet with rate signal set to unknown (127) and if it does not want to restrict the bit-rate then it does not need to update the scone packet. This lets operators deploy and manage
```
mirjak, line 244: What's the intention of this addition? Doesn't the go without saying? Why do we need to point this out separately?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -236,12 +236,12 @@ element can update the next traversing SCONE packet with the new throughput advi
 How soon the application sees the change depends on when the endpoint next sends a SCONE packet,
 since the network element cannot originate one.
 
-## Presence of SCONE Network Elements On the Data Path
+## Presence of multiple SCONE Network Elements On the Data Path
```
mirjak, line 239: I now also cover this change in PR #82

draft-ietf-scone-applicability-manageability.md
anooptomar29: By valid value I meant "not 127".  Essentially what I want to highlight is that if value is set as 127 then this restriction does not apply.

draft-ietf-scone-applicability-manageability.md
anooptomar29: This to clarify, as I get many questions from the CSPs (and their OEMs who did not participate in IETF WG discussions)  on this scenario so it is good to state it explicitly. This is an important scenario in depplyement hence I added this clarification. It may be obvious to folks who followed IETF WG discussions but it seems it is not that obvious to others.

draft-ietf-scone-applicability-manageability.md
anooptomar29: If we refer protocol doc section 7.1 below text "A network element might receive a packet that already includes a rate signal. The network element replaces the rate signal if it wishes to signal a lower value for throughput advice; otherwise, the original values are retained, preserving the signal from the network element with the lower policy. A network element MUST NOT replace a rate signal with a higher or unknown value."     This is more accurate text than what is here in appman doc.

draft-ietf-scone-applicability-manageability.md
anooptomar29: @mirjak @smishra1200 Lets continue this discussion in https://github.com/ietf-wg-scone/appman/pull/82

smishra1200:
This has been superseded by #82 and #93, both now merged. Closing this one out. @anooptomar29 the separate point you'd raised here about rate signal 127 for premium subscribers, Mirja had suggested opening its own issue for that. Want to file one if it's still worth documenting?

PR #78, opened 2026-08-03
Add two forward references

mirjak:


draft-ietf-scone-applicability-manageability.md
```diff
@@ -324,7 +325,7 @@ the QUIC flow does not exceed the throughput limits set by network policy. Alter
 can deploy SCONE purely as an advisory signal without any throttling fallback, prioritizing
 cooperative application optimization over strict compliance enforcement.
 
-## In-Band Signaling and Network Integration
+## In-Band Signaling and Network Integration {#network-integration}
```
mirjak, line 328: I would also propose to rename this to only "Network Integration" because the in-band signalling part is again only the background about the protocol but the operator relevant information is about the integration.
```suggestion
## Network Integration {#network-integration}
```

mirjak: Actually this is now cover in PR #83

smishra1200 (approved):

PR #81, opened 2026-08-03
simplify wording to communication network

mirjak:
I think this aligns better with the terminology usually used in RFCs.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -60,18 +60,18 @@ informative:
 
 --- abstract
 This document describes the Applicability and Manageability considerations for providing throughput guidance to
-application endpoints. This guidance is specifically addressed within the context of telecommunications service
-provider networks utilizing the Standard Communication with Network Elements (SCONE) protocol.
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -60,18 +60,18 @@ informative:
 
 --- abstract
 This document describes the Applicability and Manageability considerations for providing throughput guidance to
-application endpoints. This guidance is specifically addressed within the context of telecommunications service
```
anooptomar29, line 63: Lets use same terms every where.  Additionally lets use CSPs instead of operators

mirjak: We use operators everywhere. Only here this term is used once.

anooptomar29: As long as it is consistent, I am fine with operators or CSP .

anooptomar29: LGTM for merge @smishra1200 ..

smishra1200 (approved):
Looks good.

PR #82, opened 2026-08-03
editorial suggestion for multiple scone elements section

mirjak:


draft-ietf-scone-applicability-manageability.md
```diff
@@ -236,13 +236,14 @@ element can update the next traversing SCONE packet with the new throughput advi
...
+
 When multiple SCONE-capable network elements are present on the same data path, they operate
 independently, with no synchronization or control-plane coordination required between them.
-Each network element only lowers the rate signal, preserving any lower advice already set by
+This is possible because each network element is expected to only lowers the rate signal,
+preserving any lower advice already set by
 another element on the path, so the endpoint applies the most restrictive advice along the
-path (see {{Section 5.4 of I-D.ietf-scone-protocol}}). This lets operators deploy and manage
-SCONE network elements independently, without building integration between them.
+path (see {{Section 5.4 of I-D.ietf-scone-protocol}}).
```
anooptomar29, line 246: Text is protocol doc is more clear [7.1. ](https://www.ietf.org/archive/id/draft-ietf-scone-protocol-06.html#section-7.1)[Applying Throughput Advice Signals](https://www.ietf.org/archive/id/draft-ietf-scone-protocol-06.html#name-applying-throughput-advice-)   A network element might receive a packet that already includes a rate signal. The network element replaces the rate signal if it wishes to signal a lower value for throughput advice; otherwise, the original values are retained, preserving the signal from the network element with the lower policy. A network element MUST NOT replace a rate signal with a higher or unknown value

anooptomar29: So either we just refer to this section or make it clear here. We need to assume that target audience did not follow the IETF WG discussion so in appman doc we need to focus on providing more clarity even it is obvious to us.

mirjak: This text was intended to be minimal and only reference the protocol spec. However, looks like this is actually referencing the wrong section.
```suggestion
path (see {{Section 7.1 of I-D.ietf-scone-protocol}}).
```

mirjak: We should definitely not repeat a normative MUST NOT here but I guess we could make the wording a bit stronger to make it clearer, e.g. instead of "expected to" use "instructed to"...?

Or do you have a proposed wording?

anooptomar29: What about this - > When multiple SCONE-capable network elements are present on the same data path, they operate independently, with no synchronization or control-plane coordination required between them. A network element might receive a packet that already includes a rate signal set to a valid value (not unknown). The network element replaces the rate signal if it wishes to signal a lower value for throughput advice; otherwise, the original values are retained, preserving the signal from the network element with the lower policy. This way each network element, preserves any lower advice already set by another element on the path, so the endpoint applies the most restrictive advice along the path. (see {{Section 7.1 Applying Throughput Advice Signals }}).

anooptomar29: @mirjak @smishra1200 Pls take a look at the above text.  Additionally as I suggested in https://github.com/ietf-wg-scone/appman/pull/75 ,  I see a strong need to insert this below text somewhere in the doc either in this section or in some other section if there is more appropriate place   - >  " **If network element receives scone packet with rate signal set to unknown (127) and if it does not want to restrict the bit-rate then it does not need to update the scone packet"**  . Essentially this is not related to single or multiple scone n/w element scenario but this needs to be added some where to let CSP know that by not updating the SCONE pkt they can support unlimited bit-rate for their premium subscribers. This to clarify, as I get many questions from the CSPs (and their OEMs who did not participate in IETF WG discussions) on this scenario so it is good to state it explicitly. This is an important scenario in depplyement hence I added this clarification. It may be obvious to folks who followed IETF WG discussions but it seems it is not that obvious to others.

mirjak: Let's keep this PR small and editorial. I created a new PR #93 to address your proposed text. Let's continue the discussion there.

mirjak: @anooptomar29 for you second message here, I recommend to open an issue!

mirjak:
Can we merge this PR as focusing on the editorial parts only for now?

smishra1200:
Applied a small typo fix ("only lowers" -> "only lower") and merging this now. Thanks

Issue #66, opened 2026-07-28
Policies to determine throughput advice


mirjak:
The section on "Determining Throughput Constraints" has the following bullet points:

>Application-Specific Policies: Operators may set maximum bit-rates for certain types of traffic based on subscription tier or device type, for example video optimization for adaptive bitrate video, or traffic shaping for low-priority bulk transfers such as background software updates.[¶](https://ietf-wg-scone.github.io/appman/draft-ietf-scone-applicability-manageability.html#section-3.2-2.2.1)

>Dynamic Network Conditions: Constraints may be updated as network conditions change, for example when a flow moves to a different access network.[¶](https://ietf-wg-scone.github.io/appman/draft-ietf-scone-applicability-manageability.html#section-3.2-2.3.1)

>Capacity and Load Management: During periods of unusually high usage, sustained overuse, or temporary equipment faults, the network element may temporarily lower its throughput advice to manage shared capacity and guide application usage.[¶](https://ietf-wg-scone.github.io/appman/draft-ietf-scone-applicability-manageability.html#section-3.2-2.4.1)

Part of the idea of scone is actually that you don't have to play an application-specific policies. The idea is really that instead of DPI you apply the advice to all flows that indicate SCONE support. Therefore I would prefer to remove that bullet or even state that this is not recommended.

On the other two (dynamic network conditions and Capacity/Load), aren't these the same? Usually you don't change the advice only because you change cell, it rather than the new cells as a different capacity/load profile and there also has a different advice. Maybe we can remove the bullet point on dynamic network conditions and add a sentence that capacity/load might change because of network changes such as moving to a different call. Btw. changing the entire access network is not a valid example here because that would mean the network element would change as well; so there are two different independent policies applied by two different operators; this is already further discussed in the section on [Presence of SCONE Network Elements On the Data Path](https://ietf-wg-scone.github.io/appman/draft-ietf-scone-applicability-manageability.html#name-presence-of-scone-network-e).

Also I recommend adapting a title a bit. I don't think "constraints" is a right word. Maybe the word "policy" should show up instead? Like "Applying network policies to determine throughput advice"?

smishra1200:
This was addressed via PR #85, which has now merged. Closing.

smishra1200:
Reopening, apologies for the confusion above. This is tied to PR #85, which is still open, not #83. Will close once #85 actually merges.

PR #85, opened 2026-08-03
editorial suggestion for Determining Throughput Constraints section

mirjak:


draft-ietf-scone-applicability-manageability.md
mirjak, line 161: As discussed in issue #66 I also propose to merge these two points:
```suggestion
- Capacity and Load Management: Constraints may be updated as network
conditions change, e.g. during periods of unusually high usage,
sustained overuse, or temporary equipment faults. In this case, the network element
may temporarily lower its throughput advice to manage shared capacity
and guide application usage. In mobile network, this might also happen
when when a flow moves to a different cell which has a different load profile.
```

mirjak, line 152: As discussed in issue #66, I propose to remove this point as with scone one should avoid application-specific policies (only use scone as an indicator without a need to identify the application):
```suggestion
```

anooptomar29: In mobile network, this might also happen
when when a flow moves to a different cell which has a different load profile  -> **or moves to different radio access technology.**

mirjak: A different radio access technology would be switching between 4G and 5G for me. I thought switching the but staying really in the same network (same IP) is the more interesting case.

smishra1200: If I'm reading this right, Mirja, the scenario you have in mind is a cell handoff within the same radio access technology, a subscriber staying on, say, 5G the whole time, but moving between cell sites that each have their own current load. That's distinct from a RAT change (e.g. 5G to 4G), which would be a different scenario with different characteristics, and probably needs its own example rather than folding into this one.

```suggestion
- Capacity and Load Management: Constraints may be updated as network
conditions change, e.g. during periods of unusually high usage,
sustained overuse, or temporary equipment faults. In this case, the network element
may temporarily lower its throughput advice to manage shared capacity
and guide application usage. In mobile networks, this can also happen
when a flow moves to a different cell with a different load profile, even without a change in radio access technology.
```

smishra1200: Mirja, you're right on the mechanism, SCONE detection/marking is fully application-agnostic by design, a network element identifies a SCONE packet by header bits and version number only, no application-identification step anywhere in the protocol.

But I don't think that settles removing this bullet. The bullet isn't describing the SCONE network element performing DPI, it's describing an operator's policy input, what rate to advise, informed by traffic type or subscription tier, upstream of SCONE's own delivery mechanism. Once that rate is chosen, the SCONE network element delivers it to that flow the same way no matter which policy input described in this section set it. Nothing in the protocol spec says what informs an operator's chosen advice value, only that the mechanism itself must stay agnostic once a value is set. So the concept, operators may vary advice by traffic type, seems consistent with the protocol. I agree the current wording can be further refined to be more precise that this is a policy input, not a per-flow identification step the SCONE network element performs.

Suggested rewording to make that distinction explicit rather than dropping the bullet:
```suggestion
- Application-Specific Policies: Operators may set maximum bitrates
based on traffic type or device type, for example optimizing for
adaptive bitrate video, prioritizing real-time interactive traffic,
or shaping low-priority bulk transfers such as background software
updates. The SCONE network element does not identify the application
on a flow, the operator's own policy determines the advice value,
which the SCONE network element then delivers uniformly to any flow
that indicates support.
```

mirjak: subscription tier is explicitly covered by the first bullet above on "Subscriber Policies and Data Plans".

mirjak: Both of these are only examples. The switch between 4G and 5G doesn't necessarily have to do with capacity and load management though. But you proposed text is fine for me.

smishra1200: Good catch, dropped subscription tier since that's already covered by the Subscriber Policies and Data Plans bullet. Updated the suggestion above.

anooptomar29: switch between 4G and 5G  actually is more relevant here given CSPs use different bit-rate policy for its subscriber on 4G v/s 5G. Again this is directly based on inputs from CSPs. And w.r.t cell change ( inter cell handover)on same RAT is less relevant use-case for SCONE and SCONE is not meant to manage rapid cell level changes because SCONE n/w element will be hosted on PGW/UPF and there is no standard interface available that can update PGW/UPF about change of load due to change in cell. However I am ok to keep cell change but I would like to keep "moves to different radio access technology" also

anooptomar29: Pls change In mobile networks, this can also happen
when a flow moves to a different cell with a different load profile, even without a change in radio access technology -> **In mobile networks, this can also happen
when a flow moves to a different cell with a different load profile on same radio access technology or when a flow moves to a different cell on different Radio access Technology.**

anooptomar29: I will remove "The SCONE network element does not identify the application
on a flow, the operator's own policy determines the advice value,
which the SCONE network element then delivers uniformly to any flow
that indicates support"  We should not specify what CSP can not do. Some CSPs may still want to identify the application.  Essentially lets not be restrictive.

mirjak: My question is really what's the different between this point and "Dynamic Network Conditions"? If the load is temporally high that is a dynamic network condition, no? Can we maybe just merge theses two points?

anooptomar29: Key difference is what is causing change in CSPs' n/w policy. Will,  user moving from one cell to another cell with less or more load, result in CSPs to change max allowed bit-rate for the user ( based on the cell load which UPF may not even know). ?? Essentially UPF will not know Dynamic Network Conditions seen at cell level so it can not update rate signal due to cell change. UPF will know cell change but will not know load on different cells.

mirjak:
@smishra1200 @anooptomar29  I updated this PR based on our discussion! Please have a look! (Sorry the diff is somehow not great...)

draft-ietf-scone-applicability-manageability.md
```diff
@@ -147,31 +147,32 @@ per-flow context serves as the operational foundation for validating whether an
...
-updates.
-
-- Dynamic Network Conditions: Constraints may be updated as network
-conditions change, for example when a flow moves to a different access
-network.
-
-- Capacity and Load Management: During periods of unusually high usage,
-sustained overuse, or temporary equipment faults, the network element
+- Subscriber Policies and Data Plans: The throughput advice may be based
+on the subscriber's data plan. However, that does mean that the rate limit is
```
smishra1200: Fixed the logic bug here, "does mean" should be "does not mean."

```suggestion
on the subscriber's data plan. However, that does not mean that the rate limit is
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -147,31 +147,32 @@ per-flow context serves as the operational foundation for validating whether an
...
+throughput advice signal.
+
+- Capacity and Load Management: The throughput advice signal may be updated as network
+conditions change, e.g. during periods of unusually high usage,
+sustained overuse, or temporary equipment faults. In this case, the network element
 may temporarily lower its throughput advice to manage shared capacity
-and guide application usage.
+and guide application usage. In mobile network, this might also happen
+when when a flow moves to a different cell which has a different load profile,
+even without a change in radio access technology.
```
smishra1200: Per your 9/3 point, the UPF has no visibility into cell load, so removing this sentence. RAT change is already covered above.

```suggestion
and guide application usage.
```

mirjak: Yes,  but this sentence say explicitly "even without a change in RAT". So it's a different case. Also there is currently no direct signaling but still the celll change could trigger a policy update in the UPF.

smishra1200: Agree to add the point that a cell change may still trigger a policy update, even without a RAT change. Moved it up to align with the rest of the policy-based decisioning in the Subscriber Policies bullet, rather than the load-based case here.

```suggestion
and guide application usage.
```

draft-ietf-scone-applicability-manageability.md
```diff
@@ -147,31 +147,32 @@ per-flow context serves as the operational foundation for validating whether an
...
-
-- Capacity and Load Management: During periods of unusually high usage,
-sustained overuse, or temporary equipment faults, the network element
+- Subscriber Policies and Data Plans: The throughput advice may be based
+on the subscriber's data plan. However, that does not mean that the rate limit is
+static but could be updated e.g. when a
+subscriber reaches a data plan threshold or usage cap. Similarly different
+limits may apply to different access technologies and a network or mobility
+event that implies a change of access technology then leads to an updated
+throughput advice signal.
```
smishra1200: Adding the cell-change case here alongside the access-technology one, same mechanism, a mobility event triggers an updated advice signal from operator policy:

```suggestion
throughput advice signal. A mobility event that moves the flow to a
different cell with a different load profile, even without a change
in access technology, can similarly trigger an updated throughput
advice signal based on operator policy.
```

draft-ietf-scone-applicability-manageability.md
anooptomar29: LGTM

mirjak: I think this belongs actually into the Load management case here. This is due to mobility and cell load and not because of anything in the subscription.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -147,31 +147,32 @@ per-flow context serves as the operational foundation for validating whether an
...
+
+- Capacity and Load Management: The throughput advice signal may be updated as network
+conditions change, e.g. during periods of unusually high usage,
+sustained overuse, or temporary equipment faults. In this case, the network element
 may temporarily lower its throughput advice to manage shared capacity
-and guide application usage.
+and guide application usage. In mobile network, this might also happen
+when when a flow moves to a different cell which has a different load profile,
+even without a change in radio access technology.
```
smishra1200: Keeping this in Load Management per your point, fixed the "when when" typo and kept the reference to operator policy.

```suggestion
and guide application usage. A mobility event that moves the flow to a
different cell with a different load profile, even without a change in
access technology, can similarly trigger an updated throughput advice
signal based on operator policy.
```

smishra1200 (approved):
@mirjak looks good to me. Ready to merge.

smishra1200:
Hi @mirjak, PR #85 looks ready to merge. We cannot close issue #66, which this PR addresses, due to the title suggestion in that issue (renaming "Determining Throughput Constraints" to something like "Applying network policies to determine throughput advice"). Do you still want to make that change, or leave the title as is? If you do, should it go into this PR before merging, or as a separate follow-up?

mirjak:
Thanks for double-checking. Yes, let's discuss if we want to change the title and we can apply it to this PR before merging. I think a different title that has the term policy and matches the content better would be good. I'm open to suggestions!

smishra1200:
Here are some suggestions: 
- Determining Throughput Policy
- Throughput Advice Policies
- Policies for Throughput Advice

I prefer the first one as it aligns with the operator policy choices (subscriber plans, capacity and load management) rather than fixed technical constraints.

PR #86, opened 2026-08-03
Editorial clarification to network element changes section

mirjak:


draft-ietf-scone-applicability-manageability.md
```diff
@@ -244,14 +244,19 @@ another element on the path, so the endpoint applies the most restrictive advice
...
-The on-path network element can change when an application changes its access network, for
-example during QUIC connection migration or a mobility event where the IP address is unchanged.
+## Change of Network Element
+
+If the network path changes, for
+example during QUIC connection migration or a mobility event where the IP address is unchanged,
+an application flow might hit a new SCONE network element in the middle of an active connection.
 Because SCONE signaling is stateless, this transition needs no explicit teardown or state
-transfer between the old and new network elements. The endpoint and network elements follow the
-migration steps defined in {{Section 6.3 of I-D.ietf-scone-protocol}}, where the endpoint sends
```
mirjak, line 252: Removed "migration steps" because while the section in the spec talks about migration, it doesn't define a set of steps, therefore I found this confusing.

anooptomar29: This section talks about 2 scenarios 1. WiFi to cellular ( QUIC connection migration because client IP address is changed). In this case n/w element may change and new n/w element is expected to detect the flow and apply the rate limit w/o the involvement of  previous n/w element.  2. Mobility event where the IP address is unchanged . This scenario is not clear and not sure when it will happen when IP address is not changed and n/w element is changed. I think we should remove "mobility event where the IP address is unchanged" text .

draft-ietf-scone-applicability-manageability.md
```diff
@@ -244,14 +244,19 @@ another element on the path, so the endpoint applies the most restrictive advice
 path (see {{Section 5.4 of I-D.ietf-scone-protocol}}). This lets operators deploy and manage
 SCONE network elements independently, without building integration between them.
 
-## Change of Network Element During an Active Flow
-The on-path network element can change when an application changes its access network, for
-example during QUIC connection migration or a mobility event where the IP address is unchanged.
+## Change of Network Element
+
+If the network path changes, for
+example during QUIC connection migration or a mobility event where the IP address is unchanged,
```
```suggestion
example during mobility event using QUIC connection migration,
```

draft-ietf-scone-applicability-manageability.md
mirjak: Happy to remove that part. Not sure why you commented on this line of the removed text... but see my proposed commit above.

draft-ietf-scone-applicability-manageability.md
smishra1200: The on-path network element can change when an application changes its access network, for example during QUIC connection migration or a mobility event where the IP address is unchanged. Both triggers here are real and distinct. Mobile devices stay anchored to a P-GW (4G) or UPF (5G), and a standard handover, the common case, preserves that anchor and the device's IP address even though the on-path network element changes. Only a true anchor relocation changes the IP, and that's the less common, more disruptive case. So an on-path network element can change with the IP unchanged, this happens regularly in mobile deployments.

Anoop, on your comment on the mobility event with unchanged IP, see the clarification above, this is the scenario it's describing.

Mirja, your suggestion here ("mobility event using QUIC connection migration") reads differently from your own original diff, which correctly kept these as two separate triggers. This phrasing reads as if the mobility event happens via connection migration, but connection migration is the IP-changing case, and the mobility-event-with-unchanged-IP case doesn't involve migration at all. Suggested alternative to the wording above, keeping both triggers distinct:
```suggestion
example during QUIC connection migration, or a mobility event such as
a cellular handover where the device's anchored IP address stays the
same,
```

smishra1200: To be clear, please hold off on applying your suggestion above ("example during mobility event using QUIC connection migration,"). The one to apply is the wording in my reply: "example during QUIC connection migration, or a mobility event such as a cellular handover where the device's anchored IP address stays the same,"

smishra1200: Here's the full sentence with the proposed change in context, for clarity:

"If the network path changes, for example during QUIC connection migration, or a mobility event such as a cellular handover where the device's anchored IP address stays the same, an application flow might hit a new SCONE network element in the middle of an active connection."

draft-ietf-scone-applicability-manageability.md
anooptomar29: Ok I might have commented at the wrong line. But I guess you agree with my comment.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -244,14 +244,19 @@ another element on the path, so the endpoint applies the most restrictive advice
...
 Because SCONE signaling is stateless, this transition needs no explicit teardown or state
-transfer between the old and new network elements. The endpoint and network elements follow the
-migration steps defined in {{Section 6.3 of I-D.ietf-scone-protocol}}, where the endpoint sends
-SCONE packets early on the new path so a network element there can detect the flow and provide
-its own advice. If no SCONE-capable element is present on the new path, the previous advice
+transfer between the old and new network elements.
+
+As defined in {{Section 6.3 of I-D.ietf-scone-protocol}}, the endpoint is expected to send
+SCONE packets early on the new path, so the new network element can detect the flow and provide
+its own advice.
```
smishra1200: One more thing on this section, "the endpoint is expected to send SCONE packets early on the new path" reads as prescriptive language. Suggest dropping it and stating it plainly:
```suggestion
As defined in {{Section 6.3 of I-D.ietf-scone-protocol}}, the endpoint sends
SCONE packets early on the new path, so the new network element can detect the flow and provide
its own advice.
```

draft-ietf-scone-applicability-manageability.md
anooptomar29: @smishra1200 during typical hand-over when P-GW/UPF remains same  (no change in IP) but there is change in cell/eNB/gNB then SCONE n/w element would change if SCONE n/w element is hosted on eNB/gNB ( not likely). But to include this rare possibility that some CSP is going to host SCONE n/w element then also if n/w element changes then there is no way new n/w element that is hosted on eNB/gNB can detect the ongoing SCONE capable flow because app end point will not be aware of handover and unlike PGW/UPF eNB and gNB can not keep tracking every IP flow to look for SCONE pkts in DL direction. We can keep it but this is not a practical scenario. If we refer the protocol doc it talks about flow migration i.e ongoing QUIC connection getting mapped to new UDP 4 tuple , either due to change in client IP address ( wifi to celllular switching) or due to change in server IP address ( targetting to different server).

draft-ietf-scone-applicability-manageability.md
anooptomar29: LGTM

draft-ietf-scone-applicability-manageability.md
smishra1200: @anooptomar29 I was thinking of two simple scenarios, one when there is wifi/mobility or vice-versa migration (IP address changes). The second scenario for mobility that I am thinking is a more common case, that is, ordinary physical movement across cell towers. In this case, the mobile device's IP address DOES NOT change. 

The initial P-GW or UPF serving as the PDU Session Anchor (PSA) remains fixed in the data path. The PDU Session Anchor (P-GW/UPF) stays fixed and keeps the IP
unchanged; as the device moves to a new eNodeB/gNodeB, the control plane re-points the GTP-U tunnel from the old I-UPF/SGW to a new one, while traffic keeps routing back to the same anchor. So a SCONE network element sitting in that I-UPF/SGW path would see the flow migrate to a new element with no IP change, that's the common case. The network element that changes here isn't the eNB/gNB.

Please have a look at the original text and suggest what needs to change?

draft-ietf-scone-applicability-manageability.md
```diff
@@ -244,14 +244,19 @@ another element on the path, so the endpoint applies the most restrictive advice
 path (see {{Section 5.4 of I-D.ietf-scone-protocol}}). This lets operators deploy and manage
 SCONE network elements independently, without building integration between them.
 
-## Change of Network Element During an Active Flow
-The on-path network element can change when an application changes its access network, for
-example during QUIC connection migration or a mobility event where the IP address is unchanged.
+## Change of Network Element
+
+If the network path changes, for example during QUIC connection migration, or a mobility
+event such as a cellular handover where the device's anchored IP address stays the same,
```
mirjak: This is actually not a good example, because of the IP stays the same it will be the same UFP and the same network element usually.

smishra1200: Hi @mirjak, 

You make a very fair point. If SCONE is deployed centrally on the Anchor UPF (PSA), the network element indeed remains the same during standard cellular handovers. While there are 3GPP user-plane relocation scenarios where the serving element (like an Intermediate-UPF) changes while the IP remains anchored and unchanged, this represents an architectural detail that we don't need to get bogged down in here.

To keep the document precise I've simplified the opening of this section to focus strictly on the main, undisputed trigger: QUIC connection migration. Pushed as a new commit, not merged yet, please take a look. 

@anooptomar29, I also see points you were making and adding discussion on iUPF and SGW adds new level of details that best to avoid. 

I have simplified and cut out the mobility scenario. Let me know what you think?

mirjak: Looks good to me!

PR #87, opened 2026-08-03
New section on deployment modes

mirjak:
I would like to propose this new section. Understanding these deployment options will help the reader with the later sections (such as conformance monitoring) and provides clear separate guidance for all three cases.

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -289,7 +289,7 @@ guidance in {{Section 7.2 of I-D.ietf-scone-protocol}}.
 
 SCONE provides explicit signaling of throughput advice. SCONE is intended as an alternative
 to in-network rate limiting enabling more dynamic network load management and better quality of experience
-for the application user. Note that the use of SCONE throughput advise signaling can also
+for application users. Note that the use of SCONE throughput advice signaling can also
 provide these benefits to networks that are not applying rate limiting. This leads to three
 different deployment options for operators:
 
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -294,7 +294,7 @@ provide these benefits to networks that are not applying rate limiting. This lea
 different deployment options for operators:
 
 Signaling-only mode:
-: Opertors that do not enforce any rate limits might provide SCONE throughput advice to
+: Operators that do not enforce any rate limits might provide SCONE throughput advice to
   support the application with a choice of a stable rate below the capacity limit
   leading to a more smooth user experience. This operation mode does not require
   the deployment of rate limits or any conformance measurements.
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -285,6 +285,32 @@ per-flow throughput measurements in the logging infrastructure
 makes this correlation possible, aligning with the monitoring
 guidance in {{Section 7.2 of I-D.ietf-scone-protocol}}.
 
+## Deployment Options
+
+SCONE provides explicit signaling of throughput advice. SCONE is intended as an alternative
+to in-network rate limiting enabling more dynamic network load management and better quality of experience
+for the application user. Note that the use of SCONE throughput advise signaling can also
+provide these benefits to networks that are not applying rate limiting. This leads to three
+different deployment options for operators:
+
+Signaling-only mode:
+: Opertors that do not enforce any rate limits might provide SCONE throughput advice to
+  support the application with a choice of a stable rate below the capacity limit
+  leading to a more smooth user experience. This operation mode does not require
+  the deployment of rate limits or any conformance measurements.
+
+Dynamic rate enforcement mode:
+: Operators that use rate limits might disable rate enforcement for scone flows and only
+  enable it dynamically if a SCONE flow is detected to not apply the limit itself.
+  This mode saves network processing and buffer resources for compliant flows but requires
+  conformance measurements as described in the next section.
+
+Static rate enforcement mode:
+: Operators that deploy rate limit are not require to change these deployment. SCONE
+  signalling can be used in addition. In this case it is recommended to select a SCONE
+  rate that lies slightly below the actual enforcement rate in order to ensure that applications
+  can avoid to get throttled. This mode does not require conformance measurements but
+  may realise less network savings and can still lead to unwanted application impairments.
 
 ## Conformance Monitoring
 Networks that choose to provide SCONE throughput advice can implement mechanisms to


draft-ietf-scone-applicability-manageability.md
```diff
@@ -285,6 +285,32 @@ per-flow throughput measurements in the logging infrastructure
 makes this correlation possible, aligning with the monitoring
 guidance in {{Section 7.2 of I-D.ietf-scone-protocol}}.
 
+## Deployment Options
+
+SCONE provides explicit signaling of throughput advice. SCONE is intended as an alternative
+to in-network rate limiting enabling more dynamic network load management and better quality of experience
```
anooptomar29, line 291: This is not meant only for dynamic network load management , as mentioned in some other section of the doc it can also be used to enable "subscriber data plans" ( for that matter this is the most sought after user-case for CSPs at present). I think we should ensure that this section is consistent with use-case specified  in section "Determining Throughput Constraints" i.e Subscriber Policies and Data Plans, Application-Specific Policies, Capacity and Load Management, Dynamic Network Conditions.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -285,6 +285,32 @@ per-flow throughput measurements in the logging infrastructure
...
+  leading to a more smooth user experience. This operation mode does not require
+  the deployment of rate limits or any conformance measurements.
+
+Dynamic rate enforcement mode:
+: Operators that use rate limits might disable rate enforcement for scone flows and only
+  enable it dynamically if a SCONE flow is detected to not apply the limit itself.
+  This mode saves network processing and buffer resources for compliant flows but requires
+  conformance measurements as described in the next section.
+
+Static rate enforcement mode:
```
anooptomar29, line 308: I am not sure we should add this as objective is SCONE to replace the throttler. And this idea that throttler will be applied on the rate above the scone advised rates may not result in improved QoE as main drawback of throttler compared to self regulation is that throttler drops or delays the pkt w/o knowing about video delivery in the sense when more data is needed compared to when no data is needed ( bursty nature of video flows). Whole point of scone is to let app apply the rate limit instead of n/w throttler.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -285,6 +285,32 @@ per-flow throughput measurements in the logging infrastructure
...
+to in-network rate limiting enabling more dynamic network load management and better quality of experience
+for the application user. Note that the use of SCONE throughput advise signaling can also
+provide these benefits to networks that are not applying rate limiting. This leads to three
+different deployment options for operators:
+
+Signaling-only mode:
+: Opertors that do not enforce any rate limits might provide SCONE throughput advice to
+  support the application with a choice of a stable rate below the capacity limit
+  leading to a more smooth user experience. This operation mode does not require
+  the deployment of rate limits or any conformance measurements.
```
anooptomar29, line 300: I would extend this to add that CSPs can use this mode to support different bit-rates based on subscriber data plans.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -285,6 +285,32 @@ per-flow throughput measurements in the logging infrastructure
...
+## Deployment Options
+
+SCONE provides explicit signaling of throughput advice. SCONE is intended as an alternative
+to in-network rate limiting enabling more dynamic network load management and better quality of experience
+for the application user. Note that the use of SCONE throughput advise signaling can also
+provide these benefits to networks that are not applying rate limiting. This leads to three
+different deployment options for operators:
+
+Signaling-only mode:
+: Opertors that do not enforce any rate limits might provide SCONE throughput advice to
```
anooptomar29: operator -> CSP or whatever is used in the doc everywhere.

draft-ietf-scone-applicability-manageability.md
mirjak: I have an open issue about the points in the "Determining Throughput Constraints" section as I don't think these are all quite right. However, having different limits for different subscriptions is in the end just a way to manage load. I think it's better to stay high level here in this section focusing on the technology impact for an operator when deciding about the deployment mode and not the economics, because they are kind of the same no matter with mode is selected.

draft-ietf-scone-applicability-manageability.md
mirjak: No, we don't use the abbreviation CSP anywhere. we use once in the first sentence "telecommunications service provider" but I have an PR open for that. Otherwise the whole document uses the term "network operator" or just "operator"

draft-ietf-scone-applicability-manageability.md
mirjak: This is true for all modes.

draft-ietf-scone-applicability-manageability.md
mirjak: This is an operator choice and we as vendor will offer this choice because that is the deployment mode with the minimal changes. We will argue for the other two options for sure but I think it actually a good deployment strategy for operators that might not be entirely convinced yet.

Yes, this will not give the full application benefits, but that's what I wrong. I'd be okay to add another sentence and say that this deployment option is not recommended or something.

draft-ietf-scone-applicability-manageability.md
anooptomar29: No, I dont agree.  Different limits for different subscriptions does not always mean managing the load, it is due to CSPs business logic also. Lets not change the base use-cases which is one of the main motivations for CSPs to use SCONE. Lets not remove something which was added to help CSPs understand how SCONE can be useful to them and susbcriber plan based bit-rate is one of the important use-case

draft-ietf-scone-applicability-manageability.md
anooptomar29: I would remove this as this para does not serve any purpose.

draft-ietf-scone-applicability-manageability.md
anooptomar29: Ok whatever we agree to lets use that.

draft-ietf-scone-applicability-manageability.md
ihlar, line 291:
```suggestion
SCONE provides explicit signaling of throughput advice. SCONE is an alternative to using in-network rate limiting for load management or enforcement of subscription policies, providing better quality of experience
```

ihlar: I've suggested a change that should address both concerns here.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -285,6 +285,32 @@ per-flow throughput measurements in the logging infrastructure
 makes this correlation possible, aligning with the monitoring
 guidance in {{Section 7.2 of I-D.ietf-scone-protocol}}.
 
+## Deployment Options
+
+SCONE provides explicit signaling of throughput advice. SCONE is intended as an alternative
+to in-network rate limiting enabling more dynamic network load management and better quality of experience
+for the application user. Note that the use of SCONE throughput advise signaling can also
```
ihlar:
```suggestion
for application users. Note that the use of SCONE throughput advice signaling can also
```

draft-ietf-scone-applicability-manageability.md
ihlar:
```suggestion
: Operators that do not enforce any rate limits might provide SCONE throughput advice to
```

draft-ietf-scone-applicability-manageability.md
ihlar: I think it's worth mentioning this case, but we can make it even more clear that while possible, you will not get most of the benefits of SCONE. 
Maybe something like: 
"While the primary motivation of SCONE is to replace in-network rate limit enforcement, it is possible
for operators to enable SCONE in parallel to their existing enforcers. It should be noted that doing so will remove most of the positive effects that SCONE provides, both in terms of end-user QoE and for network simplifications. 
Throughput advice are intended to provide averages over at least 67 seconds, whereas typical rate-
limiting mechanisms operate on significantly shorter timescales.  An operator may attempt to mitigate issues that arise from the timescale mismatch by sending throughput advice that indicate rates that are lower than the enforced rates."

draft-ietf-scone-applicability-manageability.md
mirjak, line 351: Actually I guess this is now all covered by my new section and could be removed in we merge this in. Maybe it would be good to change the order of the section then.


```suggestion
The different deployment options are further discussed in the previous section.
```

mirjak: I don't know why it doesn't show the diff correctly but I'm proposing that everything after "Before applying rate.." could be removed.

draft-ietf-scone-applicability-manageability.md
anooptomar29: @ihlar @mirjak From actual data, co-existence of SCONE (self-regulation) and n/w throttler can sometimes result into worst performance compared to performance with only n/w throttler (no self-regulation) as self-regulation and throttler may try to limit the rate w/o any co-ordination. So I see only realistic deployment modes -
1. Self regulation through SCONE , no throttler.  Non-Conformance - Bring back throttler and stop sending  SCONE advise 
2. Rate limiter ( throttler) No scone advice

draft-ietf-scone-applicability-manageability.md
```diff
@@ -285,6 +285,32 @@ per-flow throughput measurements in the logging infrastructure
...
+Dynamic rate enforcement mode:
+: Operators that use rate limits might disable rate enforcement for scone flows and only
+  enable it dynamically if a SCONE flow is detected to not apply the limit itself.
+  This mode saves network processing and buffer resources for compliant flows but requires
+  conformance measurements as described in the next section.
+
+Static rate enforcement mode:
+: Operators that deploy rate limit are not require to change these deployment. SCONE
+  signalling can be used in addition. In this case it is recommended to select a SCONE
+  rate that lies slightly below the actual enforcement rate in order to ensure that applications
```
anooptomar29: This is really a bad combination and in this case CAP may stop doing self regulation and let rate-limiter act w/o any SCONE advice.

mirjak:
@anooptomar29 @ihlar I updates the static rate enforcement deployment option text. Please have a look!

draft-ietf-scone-applicability-manageability.md
```diff
@@ -408,7 +408,7 @@ Static rate enforcement mode:
   even when SCONE is used, this deployment is not recommended as it can risk worse performance.
   Doing so will remove most of the positive effects that SCONE provides, both
   in terms of end-user QoE and for network simplifications. This is because, SCONE enables traffic characteristic aware application driven bit-rate regulation
-  but on the other hand in-network rate limiters enforces the rate-limits arbitrarily and indiscriminately without being aware of traffic characteristics. 
+  but on the other hand in-network rate limiters enforces the rate-limits arbitrarily and indiscriminately without being aware of traffic characteristics.
```
anooptomar29: thanks Sanjay for fixing it.

anooptomar29: Thanks. Since you are planning to create a new PR on Mirja's patch, to fix some editorial. Lets also see if we can come up with better name for deployment mode 
**SCONE Signaling-only mode**: This is SCONE-only mode with no rate limiter option. This looks okay, or we may rename it to SCONE only mode 
**Dynamic rate enforcement mode:** We may need to modify the name, as ppl may confuse it with the  dynamic rate updates, which is totally different.  This deployment mode involves applying SCONE first, and then based on non-conformance by app decide to bring on rate limiter and disable scone. Let's see if we can come up with a better name. 
**Static rate enforcement mode**: We may need to modify the name, as ppl may confuse that in this mode dynamic rate updates are not supported, which is incorrect.  This deployment mode applies SCONE and the rate limiter as a guard mechanism for the worst case scenario. Let's see if we can come up with a better name.

mirjak: How about "compliance-based rate enforcement" and "Fixed rate enforcement" or "Rigid rate enforcement"?

smishra1200: How about "Conformance-based rate enforcement mode," since that reuses the existing "Conformance Measurement" term already in the doc? For the static case, maybe "Unconditional rate enforcement mode" captures it better, since the point is that enforcement applies regardless of SCONE compliance, not that the rate itself is fixed.

PR #88, opened 2026-08-06
Re submit PR#73 after rephrasing it.

anooptomar29:
What I wanted to add is that since it is allowed to the sender to decide on scone pkt frequency it can send it at lower frequency also. Essentially there are some CSPs who wants to n/w to mark every pkt received from server (dont want to have this logic in n/w element to skip some scone pkts) and to reduce the load on n/w element it is expected that sender ( server) to send pkt at lower frequency (mutually agreed).  Through this PR I added some text towards this w/o getting into specifics .

### DIFF
--- a/draft-ietf-scone-applicability-manageability.md
+++ b/draft-ietf-scone-applicability-manageability.md
@@ -210,7 +210,7 @@ application endpoint, not the network, and varies by application type and
 traffic pattern ({{Section 7.1 of I-D.ietf-scone-protocol}}, "Applying
 Throughput Advice Signals"). A network element cannot expect a predictable or
 uniform signaling cadence from the traffic itself, and instead decides its own
-update interval within that flexibility. However it is allowed to the application end point to decide on the SCONE packet sending frequency, it can send SCONE packets at lower frequency if it helps manage the network element's SCONE packet update processing load. 
+update interval within that flexibility. However it is allowed to the application end point to decide on the SCONE packet sending frequency, it can send SCONE packets at lower frequency if it helps manage the processing load on the SCONE network element.
 
 A SCONE-enabled network element updates advice in SCONE packets at least twice
 per the 67-second monitoring period ({{Section 5.2 of I-D.ietf-scone-protocol}},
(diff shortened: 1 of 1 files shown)


draft-ietf-scone-applicability-manageability.md
```diff
@@ -210,7 +210,7 @@ application endpoint, not the network, and varies by application type and
 traffic pattern ({{Section 7.1 of I-D.ietf-scone-protocol}}, "Applying
 Throughput Advice Signals"). A network element cannot expect a predictable or
 uniform signaling cadence from the traffic itself, and instead decides its own
-update interval within that flexibility.
+update interval within that flexibility. However it is allowed to the application end point to decide on the SCONE packet sending frequency, it can send SCONE packets at lower frequency if it helps manage the processing load on the SCONE network element.
```
mirjak, line 213: Sorry this still doesn't make sense because how does the application know it needs to send at a lower rate. Also, more importantly, Application should really send at a high frequency, so the network element can provide changes quickly. The only useful way to manage load it for the network element to not set the rate on all scone packets.

This is also what the spec says:

Section 7.1:
> Ideally, network elements update advice in SCONE packets at least twice per monitoring period, to match endpoint behavior

Section 8.1:
> Sending a SCONE packet every 20–30 seconds is likely sufficient to ensure that throughput advice is not lost, though endpoints might send a packet every few seconds to improve responsiveness. This period could be determined by how quickly an application is able to respond to a change in throughput advice.

mirjak: Would this help?
```suggestion
update interval within that flexibility. 

Applications are incentivized to keep the sending frequency low to avoid high overhead
but still need to manage a sending rate that enables adaption to throughput advise changes in a reasonable time. Even a high frequency is excepted to be in the order of multiple seconds as
changes to the video coding rate can usually only be applied to the next segment. There a too high
sending rate would only waste resources for both the network and the endpoints without benefit.
```

anooptomar29: This looks good. We may need to fix some nits. But overall looks good.

smishra1200 (approved):
looks good now.

PR #56, opened 2026-07-14
Issue #41, add Network Element Overload Handling section 

smishra1200:
This PR adds a new section, "Network Element Overload Handling," to address issue #41 raised by @huitema.

The section is placed after "Considerations of Processing Complexity," which establishes that SCONE reduces network
element CPU burden in the normal operating case. The new section addresses the edge case where that burden is exceeded due to an unusually high number of concurrent SCONE flows, as can occur during periods of elevated traffic or network stress.

PR #77, opened 2026-08-03
Remove "Applicability"

mirjak:
Because this document does actually talk about applicability....

PR #90, opened 2026-08-06
Adding 3rd option to mange load on n/w element due to fake flows

anooptomar29:
Suggesting a text as mentioned in my comments in the https://github.com/ietf-wg-scone/appman/pull/56. 

Operators/NE providers to identify legit flows by using existing mechanism such as SNI. This would avoid the need of NE to update the scone pkts for flows which are not legit.  This can be another option for CSPs to address this concern of processing need for large number of flows generated by attackers. Regarding the 1st point about ECH making SNI private key encrypted making it unreadable at NE. With ECH, generic SNI can still be sent in unencrypted Outer ClientHello in clear text and can be read by NE to address this concern. Essentially CSPs can agree with legit app providers on the usage of known SNI and NE can process scone pkt on flows ( on which known SNI was sent during ICH).


draft-ietf-scone-applicability-manageability.md
```diff
@@ -202,7 +202,7 @@ advice and operating without it. Which flows continue to receive
 updates and how that selection changes over time are operator
 policy decisions. Network element implementations can provide the
 configuration options through which operators define and adjust
-that policy.
+that policy. As part of 3rd option, Network element vendors and operators can identify genuine flows by using existing mechanism such as checking Server Name Indication to verify that flow is generated by genuine application endpoint. This would avoid the need of Network element to update the rate signal field in SCONE packets for flows which are not valid.
```
mirjak, line 205: I don't think we should go into that direction or provide any advise like this in this document. Identifying flow based on SNI to threat them separately is independent of scone.

draft-ietf-scone-applicability-manageability.md
```diff
@@ -184,7 +184,7 @@ period. When the number of concurrent SCONE flows grows unusually
 large, as can occur during periods of elevated traffic or network
 stress, this update obligation can become a processing concern.
```
```suggestion
SCONE processing can be stateless if all SCONE packets are always updated.
Only if the network element decide to update potentially at a lower rate or only a
selected set of SCONE-enables flows,
it might need to hold a minimum amount of per-flow state to ensure that each active flow
receives the rate signal at least once within a monitoring
period. If no rate signal is received within a monitoring period,
the rate signal is consider as expired (at least for the next monitoring period)
({{Section 5.4 of I-D.ietf-scone-protocol}}).
For stateful network element, such as those that already apply rate limiting to a selected set of flow,
the additional state and processing effort is minimal.
Network elements that were not designed to be stateful could take additional
measure to manage unexpected high load.
```

@anooptomar29 I'm proposing this update instead of your addition below. Would that help?

anooptomar29: @mirjak Looks ok to me as it no longer talks about  "When the number of concurrent SCONE flows grows unusually
large, as can occur during periods of elevated traffic or network stress, this update obligation can become a processing concern."  So I assume you are suggesting to replace the entire text under "Network Element Overload Handling" section by your update !!

draft-ietf-scone-applicability-manageability.md
```diff
@@ -184,7 +184,7 @@ period. When the number of concurrent SCONE flows grows unusually
...
 
-Network element vendors and operators have two broad options for
+Network element vendors and operators have three broad options for
```
anooptomar29: @mirjak I assume all this text will be remove and will be replaced by your update. Essentially existing option "One approach for graceful handling is
for the network element to stop updating SCONE packets for a
subset of flows entirely, rather than attempting to serve all
flows with intermittent or delayed updates. A flow that does not observe a SCONE update
for a full monitoring period will have its
throughput advice expire, causing the endpoint to operate without
SCONE-advice."  in the doc is not an option that is going to help with SCONE deployment.

draft-ietf-scone-applicability-manageability.md
anooptomar29: The reason behind suggesting this option to suggest better option than existing option in the doc to handle load on the UPF ""One approach for graceful handling is
for the network element to stop updating SCONE packets for a
subset of flows entirely, rather than attempting to serve all
flows with intermittent or delayed updates. A flow that does not observe a SCONE update
for a full monitoring period will have its
throughput advice expire, causing the endpoint to operate without
SCONE-advice."  I dont thinks CSPs are  going to deploy SCONE using this option to handle load . So I am ok to not to add SNI based app detection option to keep the number of flows under control to filter out fake flow if we remove existing option which is not practical and will not help in SCONE deployment.  FWIW, some CSPs are already considering SNI based option to detect legit app's scone capable flows. So this option is not a hypothetical option. But I agree not to mention it as it talks about SNI etc.  Can we revisit existing options in the doc and keep only those option that are realistic.

smishra1200:
@anooptomar29 @mirjak, this now merges the stateless/stateful reframing (Mirja's 8/20 suggestion) in place of the old "stop updating a subset of flows" fallback and the SNI-based option, per Anoop's 9/3 email agreeing to drop both if the old fallback also goes. 

Also fixed a truncated sentence and grammar issues from the original suggestion-apply.

Separately, this branch was 25 commits behind main (predating PR #77's citation format), so I merged main into it first to fix a stale-reference build failure unrelated to the content itself. Text is now in a clean form for review and edit. Your cawhen to merge.

mirjak:
@smishra1200 you update remove my proposed changes. However, for cleaness I created a new PR #111. Let's use that one for further discussion and close this PR.

mirjak:
Ah I now see that you added a reworked version of my text. That is fine for me as well. Sorry for duplicating this in two PRs now. However, this text was intended to also replace the first to paragraphs as well.

smishra1200:
Closing PR #90 since it is now being worked under PR #111.

